PR #35, opened 2014-08-20
Add bitrate and volume format directives

mkroman:
Added two new format directives (%BITRATE, which displays the current bitrate, and %VOLUME, which displays the volume.) to the mpd script.


dgl:
Thanks!

PR #50, opened 2014-10-06
added shadow.pl

dyladan:
This plugin is a modified version of nickcolor.pl that changes the color of the actual message instead of the color of the nick.


ailin-nemui:
at minimum you should resolve the issues mentioned at [travis report](https://travis-ci.org/irssi/scripts.irssi.org/builds/37211274#L626)

as a side note, it is possible that my [nickcolor_expando](http://anti.teamidiot.de/static/nei/*/Code/Irssi/) script can already do what you need by simply using the expando in the desired place in the /format

Issue #112, opened 2015-01-16
trackbar uses deprecated perl functions


viq:
trackbar.pl script in line 192 uses smartmatch (~~) which has been marked in newer perl as experimental, thus throwing errors and making scripts not load.

bw1:
see pull request #114

ailin-nemui:
I'd recommend avoiding this trackbar and using 1.4 -soliton or 2.2 which can be found elsewhere

GeertHauwaerts:
Updated to v1.8, it was either this or removing the script; this seemed the better option.

PR #104, opened 2015-01-02
trackbar.pl version 1.7

bw1:
the last version of
https://github.com/mjholtkamp/irssi-trackbar

PR #114, opened 2015-01-17
Issue #112

bw1:
Issue #104

### DIFF
--- a/_data/scripts.yaml
+++ b/_data/scripts.yaml
@@ -3827,11 +3827,11 @@
   contact: "peter@pfoe.be"
   description: "Shows a bar where youve last read a window"
   filename: "trackbar.pl"
-  modified: "2013-11-19 13:44:07"
+  modified: "2015-01-15 08:00"
   license: "GPLv2"
   name: "trackbar"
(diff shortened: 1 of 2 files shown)


dgl:
Please fix this so it doesn't regress the setting. Also please include a commit message that explains what the change does, not just referencing a github issue, the git history should stand alone.

Also you need to update the version in `_data/scripts.yaml`

bw1:
adding sub draw_bar

why?
in the sub cmd_mark and in the signal 'window changed' was the same code fragment.

the option 'hide_windows' only use in the signal handler. :(

and then the regex
		$config = " win1,win2 ,win3 "

	old version
		@li = split /,/,$config
		s/^\s+|\s+$//g foreach @li
		$win_name ~~ @li

	new version
		$config !~ m/$win_name/ || $win_name eq ""

	text
		is the $win_name not in the $config?

dgl:
No, this is still broken and sloppy.

If $win_name = "(status)" it will match "status" because you'll match as a regexp. Also if I'm in a channel called `#` that could be a substring of anything. 

Please also update the commit message, not just explain here.

PR #120, opened 2015-01-25
Added kban-referrals.pl

linostar:
Added kban-referrals.pl, which is a script that kickban users who post referral URLs in public messages in channels.


linostar:
Did what is suggested by dgl in his 3 comments. Waiting for your review.

PR #121, opened 2015-01-26
pangotext - A script to write various colorized text with html-tag syntax

fprintf:
This script is designed to replace the popular /rainbow script and things of that nature. This script uses tags to colorize sections of text. Example:
/pango Hello &lt;rb&gt;there is a glorious rainbow today!&lt;/rb&gt; We should go &lt;b&gt;outside&lt;/b&gt; and celebrate!

### DIFF
--- a/scripts/pangotext.pl
+++ b/scripts/pangotext.pl
@@ -8,7 +8,7 @@
 # send a rainbow-colored message where only part of the message is rainbow colored easily.
 #
 # USAGE
-# /rsay <message text here>
+# /pango <message text here>
 # NOTE: You can't put tags inside tags marked 'no inner tags' below
 # all other tags are fully nestable
 # inverse,inv   Reverse foreground and background of text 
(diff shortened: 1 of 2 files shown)


scripts/pangotext.pl
```diff
@@ -0,0 +1,147 @@
...
+# pangotext.pl - Specify color patterns in a text message using html-esque tags. In the
+#  same vein as pangotext.
+#
+# DESCRIPTION
+# The purpose of this script is to allow you to write text to the current channel with
+# complex color patterns, using simple html tag syntax. This allows you to do things like
+# send a rainbow-colored message where only part of the message is rainbow colored easily.
+#
+# USAGE
+# /rsay <message text here>
```
dequis: rsay?

fprintf: Fixed, sorry updated the comments and yaml file

fprintf:
I updated the comments and yaml file and also added a couple more features, should be ready for merge now.

fprintf:
Is this still being reviewed or is something wrong?

ailin-nemui:
@ahf ping

GeertHauwaerts:
I merged myself since we no longer need to edit the yaml file; the script is now available in the public repository.

ailin-nemui:
@fprintf You can see that it is marked with a red cross on the right side up there, which indicates that there were some problems (in this case: lack of `use strict;`) according to the current standards, your script should `use strict;`. It would be grand if you could fix this!

PR #129, opened 2015-02-01
Set record break char to newline when loading config file in people.pl

jamesoff:
I think another script of mine was breaking it, so people.pl couldn't
load its own config file as it thought the whole thing was one massive
line.


ailin-nemui:
Other scripts (cmpchans) should be fixed to properly localise $/ and it should be possible to assume \n is the value of $/ in common implicit usage...

PR #131, opened 2015-02-04
Add irssi_logger

qbit:
irssi_logger dumps chats to a PostgreSQL database for later review!


ailin-nemui:
there seems to be an unused `$search` query. maybe the feature could be implemented or otherwise removed instead?

PR #137, opened 2015-03-24
Add whocount.pl

sjktje:
Produces a user count at the end of the output of /WHO.


ailin-nemui:
I suspect this script might be confused if there are parallel WHOs going on. I'm not quite sure if this script would be better off using `redirect_event` or not

PR #138, opened 2015-04-24
Updating author details

pakerfeldt:
I've had a rename since back when this script was written. Also, the old email address is soon deprecated.


pakerfeldt:
Anyone is free to send an email to the old address for verification.

PR #140, opened 2015-04-29
Slack scripts

mortenlj:
Two useful scripts, one for Slack-users, one for anyone who has a lot of bots using privmsg in their channels.

* msg2notice - Convert privmsg to notice, if the sender is in a configurable list of nicks.
* slack_complete - When completing nicks on the Slack network, suggest mention-able nicks first (@<nick>)


mortenlj:
The build failed, but I can't figure out where in the build log it tells me what was wrong... other than that it found a critical error of some sort?

ailin-nemui:
Click on the triangle next to `$ env -u REPO_LOGIN_TOKEN ./_testing/travis/show-failures.zsh` below the test table to expand the error output.

Review [https://github.com/irssi/scripts.irssi.org/tree/gh-pages/_testing](_testing) for an overview of the test system.

In your case, the test is failing because your msg2notice script uses the deprecated 2-arg open function of perl (change it to 3-arg open)

PR #142, opened 2015-04-30
Fixed the usercount script

floriansimon1:
Used prefixes instead of Irssi variables because Irssi variables weren't working

### DIFF
--- a/scripts/usercount.pl
+++ b/scripts/usercount.pl
@@ -1,15 +1,13 @@
-use strict;
-use Irssi 20040119.2359 ();
-use vars qw($VERSION %IRSSI);
-$VERSION = "1.19";
+use Irssi 20150430.1453 ();
+$VERSION = "1.20";
 %IRSSI = (
-    authors     => 'David Leadbeater, Timo Sirainen, Georg Lukas',
+    authors     => 'David Leadbeater, Timo Sirainen, Georg Lukas, Florian Simon',
     contact     => 'dgl@dgl.cx, tss@iki.fi, georg@boerde.de',
     name        => 'usercount',
     description => 'Adds a usercount for a channel as a statusbar item',
     license     => 'GNU GPLv2 or later',
     url         => 'http://irssi.dgl.cx/',
-    changes     => 'Only show halfops if server supports them',
+    changes     => 'Using nick prefixes to count instead of incomplete Irssi variables',
 );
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
"admin" and "room owner" are ircx specific and not present in the irssi hash. Only these non-standard modes (`q`, `a`, etc) will require `prefixes` or `chr $_->{other}`. Sadly, they are not always mapped to `~` and `&`, but also to `*` or `!`. Some servers even have `-` which is less than `+`. Thus, it might be neat to compare the modes to `$server->isupport('prefix')` if the server supports `isupport`

`serverop` is quite different, it's a global server operator and *not* visible in the prefixes (`*` in `WHO`). Your change is removing this feature from the script.

c0ding:
@ailin-nemui The IRCd specific modes could be turned into settings then, couldn't they. Because having an incomplete script in to repo makes no sense to me. I've tried @floriansimon1 's changes and they actually work, which isn't the case of the initial script.

GeertHauwaerts:
@ailin-nemui close?

ailin-nemui:
@floriansimon1 it would be great if you could make the script keep the original IRCop functionality (do not touch that part of the code in function), but make it count the other prefixes. I would use a hash table for that, so that you can say  `$special{ $nick->{other} }++`. However, you need to implement the proper look-up. To do that, you can parse the result from `$server->isupport("prefix")` --  the format is `(`*`list of mode letters`*`)`*`list of characters`*. It is sorted by "power", but the only things known for sure are those actually in the standard, i,e `o` and `v`. And use the default value of `(qohv).@%+` (as required by the extended ircx standard). Then you can employ a mode letter to type table, but there is nothing really fixed to my knowledge. The actual prefixes in use vary widely, for example I found this [bug report](http://colloquy.info/project/ticket/772) on colloquy regarding to that issue. 

| letter | meaning | |
|---|-----------|----------------------------|
| u | founder | if it appears before o |
| q | owner | |
| a | admin | |
| o | op | |
| h | half-op | |
| v | voice | |
| w | member | |
| u | member | if it appears after v |
| ! | services | |

if that is too complicated for you, maybe just make it count "others" instead without going too much into detail

floriansimon1:
I can't code in Perl. My previous pull request was merely a way of contributing to your repo what worked for c0da.

PR #147, opened 2015-06-22
Cycle all channels currently connected on

SteveClement:
- Will do a /cycle in ALL currently connected channels on ALL currently connected servers
- Will exclude channels starting with: &, !, +
- The need came from a Bitlbee Race-condition where bitlbee wouldn't connect fast enough and empty channels were created. (I did not find an option to delay a connect to a Server…)

### DIFF
--- a/scripts/cycleall.pl
+++ b/scripts/cycleall.pl
@@ -1,0 +1,31 @@
+use strict;
+use vars qw($VERSION %IRSSI);
+
+use Irssi qw( servers command_bind channels);
+
+$VERSION = "0.1";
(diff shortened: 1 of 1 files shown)


ailin-nemui:
I believe your script will cycle every channel *n* times, where *n* is the number of servers you're connected to. A single loop around Irssi::channels should be enough

SteveClement:
Yes, you are right… I will implement and re-send a Pull Request

ailin-nemui:
@SteveClement  please reopen once you fixed this

PR #151, opened 2015-07-15
fnotify.pl: added support for ignoring hilights of a given priority

abair:
I use some hilights for dimming messages that I don't want to ignore completely, but this causes fnotify to send those to (in my case) growl, making them more annoying then ever.  With this patch, I can set `fnotify_ignore_hilight` to 0 and it'll ignore all hilights that I don't manually set to a higher priority.

### DIFF
--- a/scripts/fnotify.pl
+++ b/scripts/fnotify.pl
@@ -3,10 +3,11 @@ use warnings;
 use vars qw($VERSION %IRSSI);
 use Irssi;
 
-$VERSION = '0.0.5';
+$VERSION = '0.0.6';
 %IRSSI = (
 	name => 'fnotify',
-	authors => 'Thorsten Leemhuis, James Shubin, Serge van Ginderachter',
+	authors => 'Tyler Abair, Thorsten Leemhuis, James Shubin' .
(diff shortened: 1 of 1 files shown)


ailin-nemui:
commonly for the act bar this is achieved with `-actcolor %n` (stops the hilight from hilighting the number in act). Maybe such an approach would be worth implementing as well?

ailin-nemui:
if you want to use the config system, you must listen to "setup changed" signals and update your settings appropriately. Otherwise your script needs to be reloaded each time the setting is changed

ailin-nemui:
@abair please resubmit once you addressed the issues. thanks for your time and contribution

abair:
@ailin-nemui 
```quote
commonly for the act bar this is achieved with -actcolor %n (stops the hilight from hilighting the number in act). Maybe such an approach would be worth implementing as well?
```

I'm not sure I follow.  Are you suggesting I change the options available to `/hilight` instead of adding a config option in `fnotify.pl`?

Anyway, I've hooked "setup changed" and changed the ignore logic from `> $ihl` to `!= $ihl`.  I'm submitting a new pull request.

PR #153, opened 2015-07-22
added dontpart.pl script

rpowellgit:
When using irssi as a proxy, disconnecting clients like adium will disconnect irssi from irc server.  Script prevents this so you can always stay logged in to chat.

### DIFF
--- a/scripts/dontpart.pl
+++ b/scripts/dontpart.pl
@@ -1,0 +1,34 @@
+use Irssi;
+use Irssi::Irc;
+use strict;
+use vars qw($VERSION %IRSSI %config);
+$config{clientcount} = 0;
+
+
+$VERSION = "0.5";
+%IRSSI = (
+    authors     => 'ray powell',
+    contact => 'rpowell1@uchicago.edu',
+    name        => 'filter part',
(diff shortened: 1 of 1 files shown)


capi:
Doesn't this prevent any `part`, even if intentional while a proxy client is connected, from both, the `irssi` terminal _and_ a proxy client?

rpowellgit:
Yes,  but it should only block you if you still have an active proxy connection.   Once the counter hits zero, part messages come through.  Never thought about that case where i want to leave while one of my remote connections was still active. I can resubmit after I figure out how to differentiate the origin of the Part message.

capi:
@rpowellgit I always have `znc` connected to the `irssi-proxy`, so I always have a connected client, that's why I noticed. You don't have to change for me, just wanted to ask if this behavior is intentional.

ailin-nemui:
is it possible that your script also blocks other users' parting the channels?

ailin-nemui:
@rpowellgit please change your script to only stop `event part` when it is your own nick that is parting

ailin-nemui:
your script does not work for me, while the window is not closed, I still left the channel

ailin-nemui:
@rpowellgit please reopen once our concerns are addressed

PR #28, opened 2014-08-09
Add track.pl

Ziddykins:
Builds a database of users and allows you to search by nickname, ident or host. IPv6-capable and most perl regex works inside of your searches. Use '/search help' for complete information.

PR #160, opened 2015-09-19
Replaced hard-coded directory with Irssi::get_irssi_dir()

Ziddykins:
As requested. Was just easier for me to start from the beginning and create a fresh pull.

### DIFF
--- a/_data/scripts.yaml
+++ b/_data/scripts.yaml
@@ -1,4 +1,13 @@
----
+-
+  authors: 'Ziddy'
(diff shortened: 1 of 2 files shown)


dequis:
If you are adding a whole new script, "Replaced hard-coded directory with Irssi::get_irssi_dir()" does not describe this commit at all.

dequis:
For context, since it took me a bit to figure out, this is a follow up to #28

Ziddykins:
Yes, my mistake; that's correct.

PR #162, opened 2015-10-01
Add sb_search.pl

### LINKED ISSUE #161: Add sb_search.pl
sb_search.pl was removed from the default scripts because it is not here:

https://github.com/irssi/irssi/commit/a663d2f674f1408fcd54a3d1f8823dffdcc8123f

> Removed sb_search.pl which is not contributed into the script archive, I will poke coekie about it.

Which means that with my 0.8.18 git install, the only way i have to get it is through a 0.8.17 release. Also my /scrollback search broke when upgrading, surprising no absolutely one.

@coekie @GeertHauwaerts poke

Not sure why poking is needed, probably asking for permission to upload it here, which i'm not sure why it's needed since it's one of the few scripts that has a proper GPLv2 header with copyrights.

dequis:
Closes #161

### DIFF
--- a/scripts/sb_search.pl
+++ b/scripts/sb_search.pl
@@ -1,0 +1,159 @@
+# sb_search.pl - search in your scrollback, scroll to a match
+# Do /HELP SCROLLBACK for help
+
+# Copyright (C) 2008  Wouter Coekaerts <wouter@coekaerts.be>, Emanuele Giaquinta <exg@irssi.org>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


GeertHauwaerts:
Also put a copy in the script archive so it works with scriptassist (you never know if the script gets updated)

dequis:
What exactly do you mean by "script archive"?

Adding it here should trigger an update of _data/scripts.yaml from which scripts.dmp is automatically generated which should make it work with scriptassist.

GeertHauwaerts:
nm, I tought this was a request to put it in the main Irssi report with the bundled scripts :)

PR #165, opened 2015-10-18
Added xdcc_autoget userscript

fullmetal1:
xdcc_autoget allows users to store lists of xdcc bots and search terms. On a regular basis, the search terms will be fed to the bots, and their responses parsed, and new files that match each search term will be downloaded for each bot. The idea is that it automates the normally cumbersome process of finding and downloading new xdcc packs for active TV series, or other packs that the user wishes to download as soon as they are available.

### DIFF
--- a/scripts/xdcc_autoget.pl
+++ b/scripts/xdcc_autoget.pl
@@ -4,14 +4,15 @@
 # if you encounter any problems, fix it yourself you lazy bastard (or get me to), then contact me so I can add your fix and bump that version #
 # BeerWare License. Use any part you want, but buy me a beer if you ever meet me and liked this hacked together broken PoS
 # Somewhat based off of DCC multiget by Kaveh Moini.
-# USE: for help         : ag_help
-#      to run           : ag_run
-#      to halt at next  : ag_stop
-#      to set the server: ag_server 
-#      to add a bot     : ag_botadd BOT1 BOT2 *** BOTN
-#      to remove a bot  : ag_botrem BOT1 BOT2 *** BOTN
-#      to add string    : ag_add "[TEXT STRING OF TV SHOW/CHINESE CARTOON/ETC]","[ETC]",***,"[ETC]" 
-#      to remove strings: ag_rem "[TEXT STRING OF TV SHOW/CHINESE CARTOON/ETC]","[ETC]",***,"[ETC]" 
+# USE: for help             : ag_help
+#      to run               : ag_run
+#      to halt at next      : ag_stop
+#      to reset all settings: ag_reset
+#      to set the server    : ag_server 
+#      to add a bot         : ag_botadd BOT1 BOT2 *** BOTN
+#      to remove a bot      : ag_botrem BOT1 BOT2 *** BOTN
+#      to add string        : ag_add "[TEXT STRING OF TV SHOW/CHINESE CARTOON/ETC]","[ETC]",***,"[ETC]" 
+#      to remove strings    : ag_rem "[TEXT STRING OF TV SHOW/CHINESE CARTOON/ETC]","[ETC]",***,"[ETC]" 
 #      ag_next_delay                 : delay between full transfers
 #      ag_dcc_closed_retry_delay     : delay after premature transfer
 #      ag_bot_delay: delay between request and when you "SHOULD" get it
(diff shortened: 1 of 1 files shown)


ailin-nemui:
please start to improve your script firstly by addressing the warnings presented by expanding the line here in travis: https://travis-ci.org/irssi/scripts.irssi.org/builds/89707857#L767

scripts/xdcc_autoget.pl
```diff
@@ -0,0 +1,711 @@
...
+	}
+	else {Irssi::print "AG | Episodic: No";}
+	Irssi::print "AG | Data folder: $folder";
+}
+
+sub ag_initserver	#init server
+{
+	$server = " ";
+	$server = Irssi::active_server();	#keep trying to get server until it works, then continue after 5 seconds
+	if ($server !~ m/^Irssi::Irc::Server=HASH/) {Irssi::timeout_add_once(1000, sub {&ag_initserver;} , []);}
```
ailin-nemui: you can use such code instead:
```
	if (ref $server && $server->isa("Irssi::Irc::Server")) {

File: scripts/xdcc_autoget.pl [hunk 2]
```diff
@@ -0,0 +1,711 @@
...
+
+sub ag_server	#sets the current server
+{
+	$server = Irssi::active_server();	#keep trying to get server until it works
+	if ($server !~ m/^Irssi::Irc::Server=HASH/) {Irssi::timeout_add_once(1000, sub {&ag_server;} , []);}	
+}
+
+sub ag_getbots		#reads in bot list
+{
+	open(bots, "<", $botsfilename);
```
ailin-nemui: this is not wrong, just a recommendation: Modern Perl suggests to use `open(my $bots, ...` instead of bare-word file-handles.

scripts/xdcc_autoget.pl
```diff
@@ -0,0 +1,711 @@
...
+
+sub ag_search		#searches bots for packs
+{
+ 	$msgflag = 0;	#unset message flag so that ag_skip knows no important message has arrived
+	if($episodicflag)	#episodic searches are complicated
+	{
+		my $ep = sprintf("%.2d", $episode);
+		if ($format ne "" and $formatflag)
+		{
+			&ag_message("msg $bots[$botcounter] $findprefix $terms[$termcounter] $ep $format");	#first search with format
```
ailin-nemui: Please don't use `&function(parameters...)` syntax anywhere. Simply write `function(parameters...)` without `&`

scripts/xdcc_autoget.pl
```diff
@@ -0,0 +1,711 @@
...
+}
+
+sub ag_getmsg		#runs when bot sends privmsg. Avoid talking to bots to keep this from sending useless shit that breaks things
+{	
+	my $message = $_[1];
+	my $botname = $_[2];
+	$botname =~ tr/[A-Z]/[a-z]/;
+	$bots[$botcounter] =~ tr/[A-Z]/[a-z]/;
+	
+	if ($botname == $bots[$botcounter] and !$msgflag)	#if it's your bot
```
ailin-nemui: this cannot work, in Perl strings(text) must be compared with `eq`. `==` will only compare numbers (or turn text into numbers first)

scripts/xdcc_autoget.pl
```diff
@@ -0,0 +1,711 @@
...
+	@bots = ();
+	@packs = ();
+	@finished = ();
+	$termcounter = 0;
+	$botcounter = 0;	
+	$packcounter = 0;
+	$episode = 1;
+}
+
+sub ag_reset
```
ailin-nemui: this function is never used

scripts/xdcc_autoget.pl
```diff
@@ -0,0 +1,711 @@
...
+Irssi::settings_add_bool($IRSSI{'name'}, "ag_episodic", 0);
+Irssi::settings_add_str($IRSSI{'name'}, "ag_xdcc_send_prefix", "xdcc send");
+Irssi::settings_add_str($IRSSI{'name'}, "ag_xdcc_cancel_prefix", "xdcc cancel");
+Irssi::settings_add_str($IRSSI{'name'}, "ag_find_prefix", "!find");
+Irssi::settings_add_str($IRSSI{'name'}, "ag_format", "");
+Irssi::settings_add_str($IRSSI{'name'}, "ag_folder", File::HomeDir->my_home);
+
+my @totags = ();	#timeout tags (need to be purged between send requests maybe)
+
+my $nexdelay = Irssi::settings_get_int("ag_next_delay");		#delay for next pack
```
ailin-nemui: You need to make a `sub read_settings` (name does not matter) where you do all the
```
$variable = Irssi::settings_get_...
```
code. Then, you need to use
```
Irssi::signal_add('setup changed' => 'read_settings'); # do it whenever the user changes a setting with /set
read_settings(); # do it once on script load
```
Otherwise, the users of your script cannot change settings using `/set` (they would have to reload your script first, before changes take effect)

scripts/xdcc_autoget.pl
```diff
@@ -0,0 +1,711 @@
...
+	my $nexdelay = 5;
+	my $dcrdelay = 10;
+	my $botdelay = 30;
+	my $exedelay = 15;	
+	my $initflag = 1;	
+	my $sendprefix = "xdcc send";
+	my $findprefix = "!find";
+	my $folder = File::HomeDir->my_home;
+
+	Irssi::settings_get_int("ag_next_delay");
```
ailin-nemui: these lines are all useless because you do not save the setting to any variable

ailin-nemui:
I think your script currently will not work correctly with multiple connected servers, would be good to implement this in a future version

scripts/xdcc_autoget.pl
fullmetal1: I learned PERL as I wrote this script, and every single reference used bare word filehandles. I have changed the filehandles to upper case to help avoid conflicting with reserved words. I may change it in the future for a later release

fullmetal1:
I have made most of the suggested changes. Thank you for being patient with my amateurish Perl programming

fullmetal1:
I have made most of the suggested changes. Thank you for being patient with my amateurish Perl programming

ailin-nemui:
thanks for making the changes, there are a few remaining points left

scripts/xdcc_autoget.pl
fullmetal1: which one?

ag_reset now can be run by the user in IRSSI, and so can ag_stop

fullmetal1:
I tried ISA without the `=HASH` part and it failed on every attempt. In fact it only seems to work with that part added in. Instead of `$server != Irssi::active_server` shouldn't I just do `$server ne Irssi::active_server` for the same effect? Also there don't appear to be any other issues that need addressing other than that.

ailin-nemui:
Yes. Look at your original code VS the improved one:
```
if ($server !~ m/^Irssi::Irc::Server=HASH/)       |     if (ref $server && $server->isa("Irssi::Irc::Server=HASH"))
```
The original check would do something like when the server is *not* an IRC server. Now, in the changed code, we messed up that logic and changed it to whether the server *is* an IRC server. So that code would either need to say
```
unless (ref $server && $server->isa("Irssi::Irc::Server"))
```
or
```
if (!ref $server || !$server->isa("Irssi::Irc::Server"))
```
similar to ag_message. By leaving on the `=HASH`, it will always be *false*

As for the `$server != active_server`, for internal Irssi reasons you cannot compare objects using either `!=` or `eq`. It will always be false. What this code does now is, it always tries to set `$server` to the active server.
Which means the following, as far as I understand:
*  `/ag_server` is useless, because on every ag_message you will set it to the currently active server (The currently active server in Irssi depends on the server of the active window, or the one chosen with Ctrl+X in an empty window)
* Your script cannot be used reliably when connected to multiple servers
* In that case, it would in fact be easier to delete the ag_initserver and ag_server function

However I think in that case what you actually *want*, is to change the line in `ag_message` containing `$server != Irssi::active_server` to simply read:
```
if (!ref $server || !$server->isa("Irssi::Irc::Server") { return; } # give up if we don't have a server for some reason
```

scripts/xdcc_autoget.pl
ailin-nemui: fyi, this is just github being silly and not removing that old comment although you had updated the code

PR #166, opened 2015-10-19
Add at_complete.pl, for @nick style tab completion

devkev:
Twitter has a lot to answer for.

### DIFF
--- a/scripts/at_complete.pl
+++ b/scripts/at_complete.pl
@@ -1,0 +1,67 @@
+#
+# Copyright (C) 2015 by Kevin Pulo <kev@pulo.com.au>
+#
+# This adds a "@" in front of all nick tab completion, and appends ","
+# if at the start of the line.
+#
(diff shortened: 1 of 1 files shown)


devkev:
Looks like the travis job failed for transient reasons: "Couldn't resolve host 'github.com'".  I can't see any way of making it try again, I assume I don't have sufficient permissions for that.  Can someone who does have access kick it off again?  Or, if this isn't possible, should I close this PR and submit another?

xkr47:
Given that the functionality is indeed useful multiple networks, this pull request certainly proposes a better naming for the `@nick` completion than the original script `slack_complete.pl` that it is based on. However this pull request merely deletes three lines from the original script to remove the check for which network the user is currently talking in, it feels it would be better to have the functionality incorporated into the original script instead.

I have already posted a pull request #171 for supporting multiple networks in the original script, and I could easily extend it for example to support `*` as network to enable it for all networks. In fact I'll go and do it now..

PR #172, opened 2015-11-16
mh_windowfill

mh-source:


ailin-nemui:
hi,  thanks for your submission, what do you think  about this shorter script which has been used for a long time: http://anti.teamidiot.de/static/nei/Irssi/messages_bottom.pl

mh-source:
(not sure this comment will show up in the right place, but it is supposed to be an answer to ailin-nemuis comment on my pull-request)

I was unaware of that script, it largely solves the same problem, except my script also (almost) correctly deals with terminal-window resizes (and requires no manual editing of startup files)

mh-source:
My script will also move the lines at the top to the bottomline so the flow is kept intact... and it has a less restrictive license ;-)

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+
+our $windowfill_running = 0;
+
+##############################################################################
+#
+# common support functions
+#
+##############################################################################
+
+sub trim_spacelike($)
```
ailin-nemui: do not use `($)` or any other `(...)` in Perl

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+#
+##############################################################################
+
+sub windowfill($)
+{
+	my ($window) = @_;
+
+	if (ref($window) ne 'Irssi::UI::Window')
+	{
+		die();
```
ailin-nemui: `die` unloads the script, it should not be used here

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+}
+
+sub windowfill_all()
+{
+	#
+	# fill all windows with empty lines
+	#
+	for my $window (Irssi::windows())
+	{
+		if (ref($window) ne 'Irssi::UI::Window')
```
ailin-nemui: useless code, just remove it

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+	return(1);
+}
+
+Irssi::signal_add_last('mainwindow resized', 'signal_mainwindow_resized_last');
+
+sub signal_window_created_last($)
+{
+	my ($window) = @_;
+
+	if (ref($window) ne 'Irssi::UI::Window')
```
ailin-nemui: see above

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+
+	if ($windowfill_running)
+	{
+		#
+		# fill created window with empty lines
+		#
+		windowfill($window);
+	}
+
+	return(1);
```
ailin-nemui: no `return`

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+	windowfill_all();
+	$windowfill_running = 1;
+
+	return(1);
+}
+
+#
+# start script in a timeout to avoid printing before irssis "loaded script"
+#
+if (not Irssi::timeout_add_once(10, 'script_on_load', undef))
```
ailin-nemui: remove useless `if`

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+#
+# script on load
+#
+##############################################################################
+
+sub script_on_load($)
+{
+	my ($undef) = @_;
+
+	if (defined($undef))
```
ailin-nemui: useless

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+	{
+		die();
+	}
+
+	if ($windowfill_running)
```
ailin-nemui: useless

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+	{
+		die();
+	}
+
+	windowfill_all();
+	$windowfill_running = 1;
```
ailin-nemui: you can avoid this global variable by moving the Irssi::signal_add commands into this block

scripts/mh_windowfill.pl
mh-source: I use it where it works, i see no problem with it and it makes the script less suspectible to my own errors

scripts/mh_windowfill.pl
mh-source: If the script fails i want it to die isntead of continue running in a non-working state

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,244 @@
...
+#
+# script functions
+#
+##############################################################################
+
+sub windowfill($)
+{
+	my ($window) = @_;
+
+	if (ref($window) ne 'Irssi::UI::Window')
```
ailin-nemui: this code is not necessary since you are only passing windows to it

scripts/mh_windowfill.pl
mh-source: There is more useless and i have a more recent version without it, that i plan to upload of this pull request is accepted. i am not sure what you see as uselss here?

scripts/mh_windowfill.pl
mh-source: What is wrong with a return?

scripts/mh_windowfill.pl
mh-source: if timeout fails for some reason, i want the script to die()

scripts/mh_windowfill.pl
mh-source: Agreed, i have a #ifdef <PARANOID> that does a lot of what you call useless, and i call extra safetynets, i do plan on removing the safetynets later if i find the script to behave stable

scripts/mh_windowfill.pl
mh-source: I keep my souce seperated by function and use preprocessing to create the signal_add in the files where i create the signal_ subs, this might be something to fix in the future, but for now it works

ailin-nemui:
I would consider using the `$window->gui_printtext_after` call (which is fixed since irssi 0.8.17) to simply pre-pend lines at top of the window, instead of having to copy all the lines and delete the original ones

scripts/mh_windowfill.pl
ailin-nemui: I do not support protoyped Perl unless for well-documented usages such as syntax block sugar creation. You can use Google to find out up why prototypes in Perl may have been a mistake. Short-hand prototypes will also stop working once you enable 'signatures' in Perl

scripts/mh_windowfill.pl
ailin-nemui: If it is impossible for the script to fail then it must not fail

scripts/mh_windowfill.pl
ailin-nemui: Irssi::windows will only return Windows, so you do not need to make this check

scripts/mh_windowfill.pl
ailin-nemui: noone received your returned value, so you should not return one

scripts/mh_windowfill.pl
ailin-nemui: it is not possible for the timeout to fail

scripts/mh_windowfill.pl
ailin-nemui: You can upload your less useless version right away, even *before* this PR is accepted ;-)

mh-source:
I am not sure i see what you mean, can you elleborate on where gui_printtext_after should be used and what exactly it does better?

scripts/mh_windowfill.pl
ailin-nemui: if logic proves that something is not possible then you can safely put aside your paranoia in that case

ailin-nemui:
yes, it will allow you to avoid having to "copy" the original lines, so instead of having to loop over `get_lines` and `get_text`, you can simply call `$win->print_after(undef, MSGLEVEL_NEVER, "", 1) the appropriate number of times to insert a number of blank lines at top of the window, thus greatly simplifying the effort. (Sorry, I wrote the wrong name above there)

scripts/mh_windowfill.pl
mh-source: I like to have them to catch stupid mistakes in my coding, if they cause trouble i will remove them, but i see no point in removing them unless they are proven to cause issues with the script

scripts/mh_windowfill.pl
mh-source: it is not impossible for me to call windowfill() with the wrong kind of refererence, i want the script to die in that case. it will not effect the script in normal usage, again, it is a safetynet that hurts none

scripts/mh_windowfill.pl
mh-source: Again, this also protects me while i develop the script, it is not just about the offchance that irssi borks and calls something without the right arguments.

and i didnt know i could do that... however, it will not be without most of what you call useless code, the script is not live-tested enough for that yet.

scripts/mh_windowfill.pl
mh-source: What if i decide to call the function at a later date and check the return value?

scripts/mh_windowfill.pl
mh-source: famous last words :-)

scripts/mh_windowfill.pl
mh-source: But having it there doesnt hurt anyone, so i do not see the big issue?

mh-source:
I will have to look into print_after() thanks for the suggestion (i also didnt know i could make updates to the fille and make a new pull-request before the old one was accepted.. i also have a completely unused sub inthere (trim_space()) only used in some debugging that isnt there anymore - surprised you didnt cath that as "unused code" too :-))

scripts/mh_windowfill.pl
ailin-nemui: It makes the code more difficult to read and harder to review or spot more useless code, because it is full of safety nets.

ailin-nemui:
I recommend you check out the excellent `Method::Signatures` module, which allows you to write (for example)
```
method windowfill(Irssi::UI::Window $window)
```
and it will invisibly and automatically insert all the `die`s for you

mh-source:
print_after() does not work for me in irssi 0.8.15 (20100403 1617) which is the newest one available to me in ubuntus packetmanager.

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,195 @@
...
+	windowfill($window);
+}
+
+##############################################################################
+#
+# irssi command functions
+#
+##############################################################################
+
+sub command_clear
```
ailin-nemui: how about this instead:
```perl
sub command_clear
{
	my ($data, $server, $windowitem) = @_;

	&Irssi::signal_continue;
	windowfill_fill_all();
}
```

scripts/mh_windowfill.pl
```diff
@@ -0,0 +1,195 @@
...
+		$line = $window->view()->get_lines();
+
+		while ($linecount)
+		{
+			my $linetext = $line->get_text(1);
+			$line        = $line->next();
+			$window->print($linetext, MSGLEVEL_NEVER);
+			$window->view()->remove_line($line->prev());
+			$linecount--;
+		}
```
ailin-nemui: after having finished using remove_line, you *must* call `$window->view->redraw`

scripts/mh_windowfill.pl
mh-source: yes that is cleaner, i will make that change

scripts/mh_windowfill.pl
mh-source: i didnt know that. can i call it after all the lines have been removed or do i need to call it inside the loop for each line?

PR #176, opened 2015-11-21
xdcc_autoget.pl: add functionality to deal with disconnects, fix issue with removing terms

fullmetal1:
ag_rem and ag_rembot now work again. Also deals well with reconnects.


ailin-nemui:
by the way maybe you could somehow get rid of the /tmp file(s)? You can simply store everything in an `@array` and push to it instead of printing lines and reading the file

fullmetal1:
Yeah I'll get on that

Issue #144, opened 2015-05-25
Comma outputs twice


zenite:
Hey,
when I do:
/rsay something, and then some
the output is (from my side):
something,, and then some

GeertHauwaerts:
Where does the /RSAY command come from?

ailin-nemui:
either rainbow.pl or vowels.pl -- the former in this case. The bug could be in irssi as well, in case mIRC parses colour codes as always removing the trailing `,`. Someone should check that

vague666:
In rainbow.pl
        for (my $c = 0; $c < length($string); $c++) {
                my $char = substr($string, $c, 1);
                if ($char eq ' ') {
                        $newstr .= $char;
                        next;
                }
                while (($color = int(rand(scalar(@colors)))) == $last) {};
                $last = $color;
                $newstr .= "\003";
                $newstr .= sprintf("%02d", $colors[$color]);
                $newstr .= (($char eq ",") ? ",," : $char);
        }

simple test by changing ",," => "," shows no errors, not sure what ",," is supposed to accomplish, maybe meant to disable a possible background color?

No visible change in mIRC, shows double ',' when present in script, shows single ',' when script is modified

![mirc](https://cloud.githubusercontent.com/assets/8029742/11399797/e7057c88-9389-11e5-81a4-72697e4d972a.png)

ailin-nemui:
thanks for checking, sounds like an easy fix then

vague666:
Yeah, shall I push the change to the script?

ailin-nemui:
sure

dequis:
I dunno, that line looks very intentional, just removing it without understanding why it's there seems wrong.

vague666:
It does look intentional but the author should comment on it in that case, my tests show no problems with removing it from the script though.

ailin-nemui:
Thanks for cross-checking, the issue that the script creator was probably having in mind is: if you have a text like `X,00` and you insert `X<C>05,00` then the background colour would change and the comma would disappear. But it should not be an issue in the rainbow script anyway, as the script goes on and modifies the text further like `X<C>05,<C>040<C>030...`

ailin-nemui:
The other issue that he could have in mind is when colour-code parsers eat the comma after a colour code, because they believe it to be part of the colour specification. But apparently this is not the case

dequis:
Double bold?

```diff
-$newstr .= (($char eq ",") ? ",," : $char);
[...]

PR #184, opened 2015-11-25
Removed double-comma output from rainbow.pl

vague666:


dequis:
Uh, not sure about this, but i'll continue the discussion in #144

PR #185, opened 2015-11-26
mh_sbuserinfo.pl v1.00

mh-source:


ailin-nemui:
first of all please improve your pull request and commit message title by at least including the script name

mh-source:
I am not used to git/version management so all the different places i need to fill out info are a bit confusing and im never quite sure what information to put. will be sure to add full script name in the title if there is a next time

ailin-nemui:
Look for an Edit button or 🖉 icon. How does your script improve over usercount? Would it be sensible to enhance the other script instead?

mh-source:
I cant seem to find a way to edit more details than the title. but the primary reason i wrote this script was to add the current channel-limit to the statusbar, and then it will warn you if the channel usercount gets too close to the limit. I dont know exactly how it compares to "usercount"  as i havent looked at it (since it does not do what i initially set out to do, make warnings when the limit is too low)

ailin-nemui:
did you evaluate the chanfull script for that purpose?

mh-source:
I did see chanfull, but it seems to print the warning, and i wanted it on the statusbar. my script will also show a warning when the channel is opless and can be configured to only show <total users>/<limit> which is my prefered format. breaking it out into op/halfop/voice is just an added bonus of looping through the nicks. is it a big issue that my script overlaps the features of other scripts? I do think it has usefull features i didnt see any other scripts do, but i understand if you dont want too many "almost the same" scripts, just thought this one could be usefull to others too

scripts/mh_sbuserinfo.pl
```diff
@@ -0,0 +1,344 @@
...
+#
+# script functions
+#
+##############################################################################
+
+sub statusbar_redraw
+{
+	my ($channel) = @_;
+
+	if (ref($channel) eq 'Irssi::Irc::Channel')
```
ailin-nemui: why make this IRC specific? you can use this code instead:
```perl
if (ref $channel and $channel->isa("Irssi::Channel"))
```

you can make your code more readable by using "early return" pattern:
```perl
return unless ref $channel && $channel->isa('Irssi::Channel');
return unless ref $channelactive && $channelactive->{server}{tag} eq $channel->{server}{tag}
     && ...

Irssi::statusbar_items_redraw('mh_sbuserinfo');
```

scripts/mh_sbuserinfo.pl
```diff
@@ -0,0 +1,344 @@
...
+	}
+}
+
+##############################################################################
+#
+# irssi signal handlers
+#
+##############################################################################
+
+sub signal_channel_sync_last
```
ailin-nemui: You could directly bind all those events to 'statusbar_redraw'

scripts/mh_sbuserinfo.pl
```diff
@@ -0,0 +1,344 @@
...
+
+			for my $nick ($channel->nicks())
+			{
+				$users++;
+
+				if ($nick->{'op'})
+				{
+					$users_op++;
+
+				} else {
```
ailin-nemui: why not use `elsif` here and below?

scripts/mh_sbuserinfo.pl
```diff
@@ -0,0 +1,344 @@
...
+	Irssi::signal_add_last('nicklist new',         'signal_nicklist_new_last');
+	Irssi::signal_add_last('nicklist remove',      'signal_nicklist_remove_last');
+	Irssi::signal_add_last('setup changed',        'signal_setup_changed_last');
+	Irssi::signal_add_last('window changed',       'signal_window_changed_last');
+}
+
+#
+# start script in a timeout to avoid printing before irssis "loaded script"
+#
+Irssi::timeout_add_once(10, 'script_on_load', undef);
```
ailin-nemui: This is unnecessary, you aren't printing anything and there is no harm in printing before "loaded script" unless in special cases

scripts/mh_sbuserinfo.pl
mh-source: Are there other channels than irc channels? and wouldnt they behave differently (like not having limits, users or usermodes?)

ailin-nemui: There are for example SILC or XMPP channels. Users, ops and limits are a general concept. But you're also right, I am not sure if they would have the synced flag. Given that you can't be expected to go through the trouble of testing your script on all other channels maybe it's better to leave it specific.

scripts/mh_sbuserinfo.pl
mh-source: I was not aware of the elsif keyword, i will fix that

PR #186, opened 2015-11-27
add javascript filter box to script website

ailin-nemui:
and fix some html closing tags.

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,7 +32,7 @@ layout: default
 
             <dt>Is there an easy way of managing scripts?</dt> <dd>Try
             <code>scriptassist.pl</code>, it can update and manage your scripts
-            as well as install new ones and search the database.</dt>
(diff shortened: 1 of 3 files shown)


dequis:
lol travis fails because cap_sasl.pl is blacklisted

index.html
```diff
@@ -40,6 +40,7 @@
 
         <hr>
 
+        <input type="search" class="light-table-filter" data-table="sortable" placeholder="Filter" size="100">
```
dequis: Use `style="width: 100%"` instead of size="100", which just means 100 characters

dequis: Oh also `class="light-table-filter form-control"` so that it uses the bootstrap css.

dequis:
Just tested this, it's nice.

#188 addresses the travis issue.

ailin-nemui:
thanks for the comments, implemented now

PR #190, opened 2015-11-28
mh_windowfill.pl v1.04

mh-source:
fill windows so scrolling starts bottom-up instead of top-down (screenshots in source)

changes:
call windowfill* directly from signals
removed on load timeout

### DIFF
--- a/scripts/mh_userstatus.pl
+++ b/scripts/mh_userstatus.pl
@@ -1,0 +1,438 @@
+##############################################################################
+#
+# mh_userstatus.pl v1.00 (20151129)
+#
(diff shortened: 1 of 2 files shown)


ailin-nemui:
I kindly ask you not to put new submissions and updates together in one PR

mh-source:
I did not know it would happen, i thought it would create a new pull request. I was a bit confused myself when it got added here, but then i thought that was how it is supposed to work. Do you need me to redo the pull request(s) or can you work with it as-is?

mh-source:
I will close this request and make a new one

PR #192, opened 2015-11-30
 mh_userstatus.pl v1.00

mh-source:
Show in channels when users go away/back or oper/deoper

Will periodically check the channels you are on for users changing their away/oper status and print a line like:

-!- <nick> [<userhost>] is now gone
-!- <nick> [<userhost>] is now here
-!- <nick> [<userhost>] is now oper
-!- <nick> [<userhost>] is now not oper

You can also list the currently away or opered users with the commands /whoa and /whoo respectively

### DIFF
--- a/scripts/mh_userstatus.pl
+++ b/scripts/mh_userstatus.pl
@@ -1,0 +1,438 @@
+##############################################################################
+#
+# mh_userstatus.pl v1.00 (20151129)
+#
+# Copyright (c) 2015  Michael Hansen
+#
+# Permission to use, copy, modify, and distribute this software
+# for any purpose with or without fee is hereby granted, provided
+# that the above copyright notice and this permission notice
+# appear in all copies.
+#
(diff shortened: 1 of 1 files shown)


scripts/mh_userstatus.pl
```diff
@@ -0,0 +1,438 @@
...
+	'url'         => 'http://scripts.irssi.org / https://github.com/mh-source/irssi-scripts',
+);
+
+##############################################################################
+#
+# global variables
+#
+##############################################################################
+
+our $userstatus_timeouts = undef;
```
ailin-nemui: %userstatus_timeouts would be more clear here

ailin-nemui:
did you check out the anames script or the tmux nicklist?

mh-source:
the tmux script seems to be tmux specific and i never thought to check it, it does seem to check away but not serverop (and it requires tmux) and anames is a command. my script will print out the changes in a way similiar to joins and parts - as they happen.

PR #200, opened 2015-12-04
mh_hold_mode.pl v1.01

mh-source:
Emulation of ircII per-window hold_mode

changes:
nicer (imho) mh_sbmore

### DIFF
--- a/scripts/mh_hold_mode.pl
+++ b/scripts/mh_hold_mode.pl
@@ -1,0 +1,475 @@
+##############################################################################
+#
+# mh_hold_mode.pl v1.01 (20151204)
+#
+# Copyright (c) 2007, 2015  Michael Hansen
+#
+# Permission to use, copy, modify, and distribute this software
+# for any purpose with or without fee is hereby granted, provided
+# that the above copyright notice and this permission notice
+# appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+# WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL
+# THE AUTHOR BE LIABLE FOR  ANY SPECIAL, DIRECT, INDIRECT, OR
+# CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+# LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+# NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+# CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+#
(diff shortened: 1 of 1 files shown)


ailin-nemui:
just using /window scroll on/off is not enough?

mh-source:
Not if you want it "ircII"-like. it works a bit differently with regards to when it should scroll or not. i originally made this script for someone who was not happy with how irssi did its scroll, there might be others out there who also prefer the ircII way

scripts/mh_hold_mode.pl
```diff
@@ -0,0 +1,475 @@
...
+#
+# irssi signal handlers
+#
+##############################################################################
+
+sub signal_gui_key_pressed_first
+{
+	my ($keycode) = @_;
+
+	if ($keycode == $KEYCODE_ENTER)
```
ailin-nemui: does it work if you remove this check here... (see below)

scripts/mh_hold_mode.pl
```diff
@@ -0,0 +1,475 @@
...
+
+Irssi::statusbar_item_register('mh_sbmore', '', 'statusbar_more');
+Irssi::statusbars_recreate_items();
+
+statusbar_more_redraw();
+Irssi::timeout_add(1000, 'statusbar_more_redraw', undef);
+
+Irssi::command('^REDRAW');
+
+Irssi::signal_add_first('gui key pressed', 'signal_gui_key_pressed_first');
```
ailin-nemui: ...and use the signal "key send_line" instead of "gui key pressed"?

mh-source: I have just started going over it again (old script ive had lying around sincew 2007), so it was only cleaned up a little - didnt wanna mess with the internals till im a bit more into the code again. it does work as it is now, but i can look into using "key send_line" in future versions

ailin-nemui:
hard coded values for the enter key may have been broken in the git version of irssi so unless you can confirm that it works in a git build of irssi this is not acceptable

mh-source:
That is more effort than i want to put into this script. i will close the pull-request and just keep it on my own github page.

ailin-nemui:
remember this thread whenever your distribution is going to upgrade irssi in 4-5 years ;-)

mh-source:
Yeah, i will. Also warned my one known user that it might break in an upcoming version :-)  sounds a bit odd that the keycode have changed though.

ailin-nemui:
The key codes have changed to make room for one additional key to use in custom key bindings, ^J. which was previously doubling as Enter

mh-source:
Just my luck then. you wouldnt happen to know the new keycode? though i guess if i change it then it wont work in current irssi. using "key send_line" is probably a the right way forward

ailin-nemui:
yes, if possible, using keycode-agnostic code would be better

ailin-nemui:
I just submitted a PR for some scripts which makes them accept 10 and 13 to continue working with newer irssi. It's not perfect but will do for those old scripts I guess (keypad-enter key is not supported in that way)

mh-source:
I still plan on looking into "key send_line", but i can change hold_mode to accept both 10 and 13 until then and make a PR?

PR #201, opened 2015-12-04
cap_sasl 1.11

grawity:
Uploading the final version while it's still relevant :D


dequis:
People using irssi 0.8.15-0.8.17 will always exist until the heat death of the universe, so why not.

ailin-nemui:
Hi grawity, you had some thoughts/ recommendation about an irssi style API for scripting SASL, I would very much appreciate it it you could attach that to irssi/irssi#323

PR #206, opened 2015-12-09
mh_userstatus.pl v1.02

mh-source:
show in channels when users go away/back or oper/deoper

changes:
fixed issue with disabling/enabling periodical /who
/whois bug was not entirely fixed, should be now

### DIFF
--- a/scripts/mh_userstatus.pl
+++ b/scripts/mh_userstatus.pl
@@ -1,6 +1,6 @@
 ##############################################################################
 #
-# mh_userstatus.pl v1.01 (20151201)
+# mh_userstatus.pl v1.02 (20151208)
 #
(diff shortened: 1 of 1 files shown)


ailin-nemui:
what actually is this whois bug and what causes it?

mh-source:
irssi updates away/oper when you do whois, and generates a "nicklist * changed" so it would spam the channel(s) with "X is now here" then "X is now gone" every time you do /whois on a nick you share channel with. thats why i setup the global variable to catch when we do /whois and ignore the events. the reason it was not fully fixed in the earlier fix was that i did not catch the whois numeric signal early enough (signal_register() instead of signal_register_first())

ailin-nemui:
ok cool

PR #210, opened 2015-12-14
fix bug where attempt to download previously downloaded and deleted pack caused abberant behaviour

fullmetal1:
Irssi keeps saying 

```perl
Irssi: critical irssi_bless_iobject: assertion '(type & ~0xffff) == 0' failed"
```

### DIFF
--- a/scripts/xdcc_autoget.pl
+++ b/scripts/xdcc_autoget.pl
@@ -297,7 +297,7 @@ sub ag_packrequest	#sends the xdcc send request, and retries on failure
 	if (!$reqpackflag)
 	{
 		$reqpackflag = 1;
-		Irssi::signal_add("dcc get receive", "ag_opendcc");		#init DCC recieve init flag
+		Irssi::signal_add("dcc created", "ag_opendcc");		#init DCC recieve init flag
 		$statusbarmessage = "Getting pack #$packs[$packcounter] on $bots[$botcounter]";
 		ag_message("msg $bots[$botcounter] $sendprefix $packs[$packcounter]");
 		push(@totags, Irssi::timeout_add_once($botdelay * 1000, sub { if (!$downloadflag) { &ag_packrequest; } } , []));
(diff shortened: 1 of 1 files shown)


fullmetal1:
Also there's a small chance that irssi will crash the first time that it would print out that message (which seems to happen when the command '/dcc close get [BOT]' is run

ailin-nemui:
can you give more detail how the bug happens? maybe this is also a bug in irssi itself which should be fixed

fullmetal1:
From what I can tell it seems to happen when $server->command("dcc close get $bots[$botcounter]");  is called.

I've managed to find a relevent bug report https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=740003

As far as I can tell the issue is that the '/DCC' command shouldn't be sent to the server, however, sending the command to the active window (or any window) seems to have the same effect. The command itself can be replaced with DCC::destroy(), but DCC::destroy() doesn't seem to exist as a module in Irssi::*.

I'm on v0.8.15 if that makes a difference

ailin-nemui:
that relevant bug which you found is caused by the $server which you were using, getting disconnected. Such could happen during netsplits or manual /disconnect for various reasons that disconnected you. Even if Irssi reconnects, the $server will not remain valid if it was stored in a global variable. Is that possible to have happened to you?

Due to some limitation in Irssi API, you should actually never store Irssi::active_server or any other $server or any other "Irssi object" in a global variable. It is better to always get a fresh $server local to the function where you are using it, for example like this:
```perl
if (my $server = Irssi::server_find_tag("mynetworktag")) {
    $server->command(...);
} else {
   # not connected to mynetwork
}
```
Failure to do so may even lead to Irssi crashing ;-(

If you want your script to be stuck to a specific server, remembering that server's "tag" is safe. E.g. `$global_server = Irssi::active_server->{tag}`

You can send global commands that do not require a server with `Irssi::command(...)`

To close a DCC it may work to call `$dcc->close` but I have not tested that. You get $dcc from any [signal](https://github.com/irssi/irssi/blob/master/docs/signals.txt#L183) which lists DCC_REC, Example.
```perl
sub dcc_handler {
    my ($gdcc, ...) = @_;
    $gdcc->close;
}
```

fullmetal1:
I have managed to reproduce the issue.

Loading this script
'''perl
sub test
{
	my ($gdcc) = @_;
	$gdcc->close;
}
Irssi::signal_add("dcc created", "test");
'''
and trying to get a dcc file will trigger the effect

I guess I'll also open a bug report. I guess this pull request shouldn't be merged until the issue is sorted out

ailin-nemui:
thanks for digging in to this

ailin-nemui:
Can you try if this code behaves stably:

* Use signal "dcc request" instead of "dcc created".
* As the first thing in your signal handler (ag_opendcc), call `&Irssi::signal_continue;`

fullmetal1:
That results in
>19:20 -!- Irssi: critical irssi_bless_iobject: assertion '(type & ~0xffff) == 0' failed
19:20 Use of uninitialized value in transliteration (tr///) at /home/alex/.irssi/scripts/autorun/xdcc_autoget.pl line 426.
19:20 
19:20 Use of uninitialized value $botname in string eq at /home/alex/.irssi/scripts/autorun/xdcc_autoget.pl line 429.
19:20 
19:20 -!- Irssi: critical irssi_bless_iobject: assertion '(type & ~0xffff) == 0' failed
19:20 -!- Irssi: critical g_source_remove: aSegmentation fault

When the active window is the servers, and, inexplicitely, seems to kill my script by running ag_stop 
¯\\\_(ツ)_/¯

fullmetal1:
It works 
¯\\\_(ツ)_/¯

PR #214, opened 2015-12-20
mh_windowfill.pl v1.06

mh-source:
fill windows so scrolling starts bottom-up instead of top-down (screenshots in source)

changes:
added changed field to irssi header


mh-source:
I wonder why this have neither been accepted nor commented on since dec 20th. did it slip through the cracks?

ailin-nemui:
I'll merge it now

PR #218, opened 2016-01-12
Add autonickprefix.pl sscipt

Juerd:


Juerd:
Note: build started before 'use strict;' was added, but didn't restart after it was added. Probably safe to merge.

ailin-nemui:
hi, you are missing a `use Irssi::TextUI` before `use Irssi`. This is required for users of the modular Perl support in Irssi, otherwise the `gui_*` functions won't be exposed. You can click on the triangle on line 795 of https://travis-ci.org/irssi/scripts.irssi.org/builds/101852040 to see the error message(s)

PR #219, opened 2016-01-15
xdcc_autoget v2.0 :) Now does all bots concurrently

fullmetal1:


fullmetal1:
Just noticed a bug. Please don't pull yet

fullmetal1:
Working now

scripts/xdcc_autoget.pl
```diff
@@ -200,6 +203,59 @@ sub ag_getfinished		#reads in finished packs list
 	chomp(@finished);
 	@finished = ag_uniq(@finished);
 	close(FINISHED);
+	&ag_clearcache;
+	@finished = &ag_uniq(@finished);
```
ailin-nemui: do you need to run ag_uniq  2 times here :-)? also please without `&` (& = perl4 style)

scripts/xdcc_autoget.pl
```diff
@@ -211,50 +267,34 @@ sub ag_clearcache		#clears cache of saved packs
...
 sub ag_remtimeouts	#remove timeouts to avoid multiple instances of everything
 {
-	foreach my $to (@totags)	#remove timeouts and clear array
+	my($botcounter) = @_;
+	foreach my $to (@{$totags[$botcounter]})	#remove timeouts and clear array
 	{
 		Irssi::timeout_remove($to);
 	}
-	@totags = ();
+	$totags[$botcounter] = ();
```
ailin-nemui: I think `@{$totags[$botcounter]} = ()` would be more correct

scripts/xdcc_autoget.pl
```diff
@@ -315,109 +366,104 @@ sub ag_opendcc	#runs on DCC recieve init
...
-		&ag_remtimeouts;
-		$dccflag = 0;
-		$downloadflag = 1;
-		$statusbarmessage = "Downloading $filename from $botname, #$packs[$packcounter]"; 
-		foreach my $n (@finished)		#don't redownload finished packs
+		$bot =~ tr/[A-Z]/[a-z]/;
+		if ($botname eq $bot and !$filedownloadflag)	#if it's our bot and the file is not already being downloaded, let user know, and stop any further AG pack requests until finished
 		{
-			if ($n eq $gdcc->{'arg'})	#if file already downloaded, cancel
+			&ag_remtimeouts($botcounter);	#stop any other skips
```
ailin-nemui: please remove the `&` from your head ^^

fullmetal1:
The repeated @ag_uniq snuck in while I was furiously pasting in code from another function.

Missed the @{} casting

I got rid of all of the ampersands then only added back in the ones that Irssi's perl interpreter bitched about. That's an acceptable programming practice yes?

ailin-nemui:
no, the only acceptable programming practice is understanding what two things `&` does. so that you can use it properly. You've introduced a new bug by simply removing it everywhere. Let me try to explain you:

`&function($param1, $param2)` usually does not need to be written. You should just write `function($param1, $param2)` instead.

Sometimes functions expect a hash or array, so they would normally be called like `function(%myhash)`, but if you know what you are doing and have your reasons you may need to pass in something differently, then you can **overwrite the prototype** by calling `&function($hashreference)`. You do not usually need to do this.

If your function does not require any parameter, you may be tempted to write `&function;`. In this case the recommended way to call your function should be: `function();`

Now, when do we **need** to write `&function;`? It is a special case that passes on the current `@_` (the current function parameters) to the next function. This is always the right way to call `&Irssi::signal_continue;` (if you do not want to change the signal parameters) since you want to continue the signal with the original parameters. You *must* call `&Irssi::signal_continue;` or `Irssi::signal_continue(@_)` or `signal_continue` with a list of changed signal parameters.

Also, if we need a reference to a function, then we need to write `\&`, as in `$code = \&function;`

Finally, *why* does Perl sometimes accept `function;` instead of `function();` ? The reason is that if it has already seen `sub function` somewhere **before** you call `function;`, it knows that you want to call this function and you can be lazy and omit the `()`. I prefer to explicitly write `function();` always.

scripts/xdcc_autoget.pl
```diff
@@ -262,162 +301,168 @@ sub ag_getmsg		#runs when bot sends privmsg. Avoid talking to bots to keep this
...
+		$reqpackflag[$botcounter] = 1;
+		ag_message("msg $bots[$botcounter] $sendprefix $packs[$botcounter][$packcounter[$botcounter]]");
+		push(@{$totags[$botcounter]}, Irssi::timeout_add_once($botdelay * 1000, sub { if (!$downloadflag[$botcounter]) { ag_packrequest($botcounter); } } , []));
 	}
 }
 
 sub ag_opendcc	#runs on DCC recieve init
 {
-	&Irssi::signal_continue;
+	Irssi::signal_continue;
```
ailin-nemui: at least here the `&` must be used

scripts/xdcc_autoget.pl
```diff
@@ -262,162 +301,168 @@ sub ag_getmsg		#runs when bot sends privmsg. Avoid talking to bots to keep this
...
 }
 
 sub ag_skip
 {
-#	Irssi::print "AG | SKIP $msgflag $episodicflag $episode $#packs $packcounter $#terms $termcounter $#bots $botcounter"; 
-	&ag_remtimeouts;	#stop any other skips
-	$reqpackflag = 0;		#allow pack requests now that transfer is finished
+	my($botcounter) = @_;
+	my @packlist = ();
+	try {my @packlist = @{$packs[$botcounter]};}	#workaround for @{} being dumb if an array inside of an array is empty
```
ailin-nemui: another common code which also looks ugly and could be used instead of try/catch: 
`@{$packs[$botcounter]//[]}`
that translates to: if `$packs[$botcounter]` is undef, use `[]` instead (which is an anonymous empty array reference)

PR #220, opened 2016-01-20
Fix escaping for active_notice.pl when a line starts with a percent sign

nvisser:
Fixes notices appearing with markup if the line starts with a % sign. For example, a line beginning with %I would appear inversed.


ailin-nemui:
thanks

PR #225, opened 2016-01-24
mh_userstatus.pl v1.04

mh-source:
show in channels when users go away/back or oper/deoper

changes:
added changed field to irssi header


ailin-nemui:
I think it's better to submit actual changes ;-)

mh-source:
I agree, and i have a few small updates like that, kinda sorry how im flooding you with them. but id also like to keep irssi.org in sync with my own repository

PR #229, opened 2016-01-29
aspell_complete: Fix crash when the dictionary is unset

dequis:
This will be fixed on the irssi side too.


dequis:
Irssi fix: https://github.com/irssi/irssi/pull/413

PR #230, opened 2016-01-31
logcompress using IO::Compress::GZip

vague666:


jillest:
Note that, unlike logcompress.pl, this performs the compression in the irssi process, potentially blocking it for a long time if the file is large.

For small files it is certainly an improvement, though.

vague666:
I'll look at forking it for sure, thanks for the pointer Jilles

2016-02-05 23:57 GMT+01:00 Jilles Tjoelker <notifications@github.com>:

PR #231, opened 2016-02-03
update scriptassist.pl

ailin-nemui:
* remove prototypes
* fix recently introduced 304 error issue
* comment out rating commands (service is down since 12 years)
* rewrite deprecated hash checks
* add name translation (best effort)
* do not fail if .pl is given in install command
* use more 'strict' and less bareword filehandles
* pass through /help
* change default sources to https://

### DIFF
--- a/scripts/scriptassist.pl
+++ b/scripts/scriptassist.pl
@@ -5,21 +5,19 @@
 
 use strict;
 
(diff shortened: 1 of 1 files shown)


dequis:
Specifying what was changed in the commit message would be nice

PR #236, opened 2016-02-08
Fix appearance / user formatting of awaybar

dreamcat4:
Source: https://github.com/ronilaukkarinen/weed

### DIFF
--- a/scripts/awaybar.pl
+++ b/scripts/awaybar.pl
@@ -1,6 +1,19 @@
 # awaybar.pl -- initially built for Irssi 0.8.9
-# thanks to mood.pl for practically allowing me
-# to copy the approach..
+# thanks to mood.pl for practically allowing me to copy the approach..
+#
+# Usage:
+#
+# Creating the away bar
+# This assumes you have adv_windowlist, and that we add before it's window named 'awl_0'
+# /set awl_shared_sbar RIGHT
+# /statusbar window add -after lag -priority 10 -alignment right awl_shared
+# /sbar awl_shared add -before awl_0 -alignment right awaybar
+#
+# Appearance
+# To set the look and feel of the awaybar, and save it to your .theme file:
+# /format sb_awaybar "%1%Waway: ${0-}"
+# /save
+
 use strict;
 use Irssi;
 use Irssi::TextUI;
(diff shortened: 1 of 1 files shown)


dequis:
I'm not sure I understand this change properly.

What about users that don't have AWL?

What about users that upgrade this script through, say, scriptassist, and don't have that format entry?

dreamcat4:
Hi @dequis. Thanks for looking at this PR.

1. Yes that is a good point and sorry that my instruction is so limited like that. It is due to my own lack of knowledge / in-experience. As I simply don't know enough about statusbars. That instruction is cpoed from the instructions / documentation of the Weed theme (link provided in commit msg).

Yes it would certainly benefit to make a better or more generic instruction, including for irssi's standard away bar. The help text is just a supplementary usage infos which is fine to reject entirely if you wish. Just thought it would be better / more helpful than nothing at all.

2. Yes that situation is fine. I have tested without the variable set. So that is same thing for any user who was upgrading from previous version. They will see no difference in the output. The previous version prints / outputs an unformatted text. So will this version if the variable isn't user-set. Nothing is getting broken with that. AFAIKT the scripts original author seemed to try make in the commented line a specific theme setting string, but he gave up on it. So then this PR fixes that problem and explains to the user with an example how they can customize it.

ailin-nemui:
There are some problems:
- you cannot change abstracts using /format, so the instruction you added is wrong about that
- as I explained to you on IRC your awl_shared_sbar setting is wrong
- the /sbar command is also incorrect (should be awl_shared instead of awl_0)
- I don't think this adv_windowlist specific instructions belong in that script
- Also, previously the script would show "Away: ...", with your change it will *only* show the away reason
nevertheless I think it would be fine to make the format configurable in general

dreamcat4:
> nevertheless I think it would be fine to make the format configurable in general

Well great. That was the main thing.

> you cannot change abstracts using /format, so the instruction you added is wrong about that

OK so now I'm not entirely sure how to achieve the new user setting since you said that. If:

```
/format sb_awaybar "%1%Waway: ${0-}"
```

is wrong.

Well no hurry. One of us is bound to come back and figure it out eventually.

ailin-nemui:
I will close this misleading PR, please reopen if you can fix the issues mentioned in here

PR #238, opened 2016-02-10
Support ##channels, &bitlbee, *znc_module

dreamcat4:
This improves the regex matching for other kinds of open windows / channels.

### DIFF
--- a/scripts196/go.pl
+++ b/scripts196/go.pl
@@ -47,7 +47,7 @@ sub cmd_go
 	$chan =~ s/ *//g;
 	foreach my $w (Irssi::windows) {
 		my $name = $w->get_active_name();
-		if ($name =~ /^#?\Q${chan}\E/) {
+		if ($name =~ /^(##?|\&|\*)?\Q${chan}\E/) {
 			$w->set_active();
 			return;
 		}


dequis:
Feels like there should be a better way

dreamcat4:
@dequis Yeah maybe. But what to you mean by that? You could mean some configurable list of prefix strings. But I think maybe you mean 'just any non- alpha numeric prefix characters'...

eg `^[^0-9a-zA-Z]*`

ailin-nemui:
we should increase the version if we make a change so people get upgrade notification

ailin-nemui:
@dreamcat4

dreamcat4:
Just waiting for @dequis to clearify what he meant / how to take further.

dequis:
The valid channel prefixes are network specific and defined in the CHANTYPES token of RPL_ISUPPORT. For example, ircnet has CHANTYPES=#&!+

The second # in ## channels isn't a prefix, it's part of the name, and the way it's used in freenode is just a convention. Channel names can have any 8bit chars except null, cr, lf, space, and maybe some other one i'm missing, and some irc servers enforce even more restrictions on top of that. But it's perfectly legal to have `#$%^&$^$%^` or even names with no ascii alphanumeric characters at all. A regexp that skips all of that would be wrong.

This patch solves your immediate needs (and it's certainly better than working for # and breaking for everything else) but it still fails to work with some cases, and a regexp doesn't seem like the cleverest way to do this. Special-casing the double ## of freenode seems fair though.

Also, not sure if i'm misreading the diff, but did you just move the script to a new directory?

ailin-nemui:
@dreamcat4 please move the script back to the correct directory and increase the version number if you want this simple PR accepted

ailin-nemui:
feel free to reopen once issues are addressed

PR #245, opened 2016-03-15
Submit my collection of scripts from my dotfiles.git

avar:
These are all irssi scripts that I maintain & use myself that I think
are useful for general use, none of this has anything hardcoded related
to me and when applicable has a configurable interface.

### DIFF
--- a/scripts/bitlbee_hide_password.pl
+++ b/scripts/bitlbee_hide_password.pl
@@ -1,0 +1,64 @@
+use strict;
+use warnings;
+use Irssi;
+
+our $VERSION = '1.0';
+our %IRSSI = (
+    authors     => 'Ævar Arnfjörð Bjarmason',
+    contact     => 'avarab@gmail.com',
+    name        => 'bitlbee_hide_password.pl',
+    description => 'Hide your REGISTER and IDENTIFY password lines in &bitlbee from screen & logs ',
+    license     => 'Public Domain',
+    url         => 'http://scripts.irssi.org & https://github.com/avar/dotfiles/blob/master/.irssi/scripts/bitlbee_hide_password.pl',
+);
+
+# HOWTO:
+#
+#   /load bitlbee_hide_password.pl
+#
+# When you're using Bitlbee and do either:
+#
+#    REGISTER <password>
+#
+# or:
+#
+#    IDENTIFY <password>
+#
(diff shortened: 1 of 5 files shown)


dequis:
>bitlbee_hide_password.pl

You could just use /^msg (which sends a message without printing it). Or server password. Or sasl.

avar:
@dequis Yeah I could do that, there's also "/OPER" which does a similar thing, but the reason I have that script is that just typing `identify` in the window is burned into my muscle memory, the script is also handy for anyone else used to doing the same.

scripts/bitlbee_hide_password.pl
```diff
@@ -0,0 +1,72 @@
...
+    # We only care about the &bitlbee control channel
+    my $target = $tdest->{target};
+    return unless $target eq '&bitlbee';
+
+    # We're matching against $stripped but replacing both because the
+    # $data thing is escaped and much harder to match against.
+    #
+    # We're just replacing nick mentions, so e.g. if you say "Hi I'm
+    # bob here but my username is bobby" it won't turn into "Hi I'm
+    # bobby here but my username is bobby".
```
ailin-nemui: I think you c&pd the comments but now they're in the wrong script and sort of confusing

scripts/munge_own_nickname_to_username.pl
```diff
@@ -0,0 +1,74 @@
...
+    # $data thing is escaped and much harder to match against.
+    #
+    # We're just replacing nick mentions, so e.g. if you say "Hi I'm
+    # bob here but my username is bobby" it won't turn into "Hi I'm
+    # bobby here but my username is bobby".
+    #
+    # The illusion here isn't complete, e.g. if you do /NAMES your
+    # nick will show up and not your username, but I consider that a
+    # feature.
+    if ($stripped =~ /^<.?\Q$server_nick\E> /s) {
```
ailin-nemui: your script only works when the theme uses <nick> format, but it's difficult to do it better in currents' irssi

avar:
@ailin-nemui Thanks a lot for the feedback. I fixed the confusing comment (and fixed the s/hawe/have/ spelling error there).

The second issue with the nick matching regex is trickier as you say, for now I've just added documentation in the comment alerting users to the  issue.

PR #247, opened 2016-03-20
new script betterlist.pl

hopkiw:
this script essentially applies grep.pl to LIST command output; it seemed like it might be more work to amend that script, so locally i've aliased list to this command which allows me to search list output


ailin-nemui:
if the user gives an invalid regex your script may crash(?)

PR #256, opened 2016-04-18
support filling the javascript filter from url hash

ailin-nemui:
[skip ci]


ailin-nemui:
with that we can do https://scripts.irssi.org/#q=sasl

PR #257, opened 2016-04-18
autorun scripts in the autorun directory

vague666:
Uses Inotify to watch for updates in the autorun directory and takes appropriate action


dequis:
`%IRSSI`?

PR #259, opened 2016-04-26
Add hilightcmd with proper shell string quoting

tslocum:


scripts/hilightcmd.pl
```diff
@@ -0,0 +1,50 @@
...
+    if (Irssi::settings_get_bool('hilightcmd_privmsg')) {
+        $opt = MSGLEVEL_HILIGHT|MSGLEVEL_MSGS;
+    }
+
+    if (($dest->{level} & ($opt))
+	&& ($dest->{level} & MSGLEVEL_NOHILIGHT) == 0
+	&& (Irssi::active_win()->{refnum} != $dest->{window}->{refnum}
+            || Irssi::settings_get_bool('hilightcmd_currentwin'))) {
+
+        $stripped =~ s/"/\\"/g;
```
ailin-nemui: I believe you can omit this replacement. did you test it?

tslocum: Thanks for pointing this out.  I've updated this line to trim whitespace instead.

PR #263, opened 2016-04-26
Add script beforespace

ailin-nemui:
Rebind certain keys so that they are inserted before the space


dequis:
I don't get it. Is this like some sort of autocorrect?

ailin-nemui:
If you type "hello ," it will turn into "hello, "
Also useful with tab completion and continued typing

PR #266, opened 2016-04-26
Add settingshelp

ailin-nemui:
A copy of https://irssi.org/documentation/settings/ as /help script


dequis:
This really should be a built-in feature

ailin-nemui:
so whaddowedo

dequis:
We... add it as a built-in feature? But later. I think it's okay to have this script for now (and for released versions)

PR #281, opened 2016-06-15
Updated tinyurl.pl to work with recent perl and tinyurl.com versions

efuce:
Updated tinyurl.pl to work with more recent versions of perl. Fixed the url matching regex to work with the current layout of tinyurl.com.


ailin-nemui:
Maybe use Www::Shorten::TinyUrl instead?

efuce:
Thanks for the tip. I did not known about WWW::Shorten::TinyUrl. Updated the pull request.

ailin-nemui:
btw you do not need to update scripts.yaml

efuce:
Thanks for informing me. I will remember that for any future commits.

PR #282, opened 2016-06-17
Add script buffer.pl

pierrot14:


scripts/buffer.pl
```diff
@@ -0,0 +1,560 @@
+use strict;
+use warnings;
+
+use vars qw($VERSION %IRSSI);
```
ailin-nemui: You do not need this if you use `our`

pierrot14: Thanks Nei. I prefer `our` to `use vars`. I probably copied and pasted those first lines from an old script.

I removed the line containing `use vars`.

PR #287, opened 2016-06-28
Line 37-45 missing OS if/else for cross platform

diqidoq:
#### Line 37
 + @TODO make this command system undepended or provide user info for required alias
 + Debian version: system("iceweasel --new-tab $lasturl &>/dev/null"); tmp added.

### DIFF
--- a/scripts/urlgrab.pl
+++ b/scripts/urlgrab.pl
@@ -38,7 +38,7 @@ sub url_cmd{
    # Debian version: system("iceweasel --new-tab $lasturl &>/dev/null"); tmp added.
    my $osname = $^O;
    if($osname eq 'linux'){
-     system("iceweasel --new-tab $lasturl &>/dev/null");
+     system("firefox --new-tab $lasturl &>/dev/null");
    } else {
      system("netscape-remote -remote 'openURL($lasturl)' &>/dev/null");
    }
--- a/scripts/urlgrab.pl
+++ b/scripts/urlgrab.pl
@@ -34,7 +34,14 @@ sub url_cmd{
 	  Irssi::print("No url captured yet");
 	  return;
    }
-   system("netscape-remote -remote 'openURL($lasturl)' &>/dev/null");
+   # @TODO make this command system undepended or provide user info for required alias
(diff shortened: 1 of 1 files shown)


dequis:
Something makes me think you didn't exactly aim for portability here.

diqidoq:
Yeah right @dequis , that's why I suggest a cross platform switch, when there was nothing than a windows `netscape-remote ` in there before ... *facepalm* ... define your meaning of portability for me. Works fine here now on Windows and Debian.

dequis:
Sorry for that, it's true that it's better than before

"netscape-remote" isn't a windows thing, it's just that the script hasn't been changed significantly 2002. I don't think anyone actually cares about windows.

Does "firefox" launch "iceweasel" in debian? If not you might want to go for something more generic, like "xdg-open"

diqidoq:
Oh, ok, sorry for that here too. Now I know what you are after. xdg-open is a good and "more open" solution. True. But, sadly it only works on http:// urls, not on short url.tld urls. Under Debian and Ubuntu there is no --url prefix or parameter flag available for xdg-open. The command would be xdg-open http://you-url.tld, which misses the new tab option to be sure that we don't fade away from actual open browser session of user and we are not solid with non http:// urls.

Otherwise I didn't have tested yet (no time today no more, need to sleep, its in the morning) if the script catches non http:// urls anyway?

To your question: Iceweasel is just another _branding_ for Firefox. It's actually the same. For some years there were some trouble between Debian and Firefox regarding security concerns of Debian, which prevented Debian from providing latest Firefox updates in the Debian package repositories. This has caused that Firefox forced them legally to rename this package. But this solved and Debian goes back slidely to Firefox now. You can use both commands the same way, not only in Terminal, also in Run console or any other scenario like Destop icon files, etc.

ailin-nemui:
maybe the best way would be to make this a setting so users can configure their favourite opening program? obviously the script is rather dated, netscape, hehehe. And maybe detect if the URL contains :// and if not, prefix it with http://

diqidoq:
awesome work guys! thanks for chiming in.

PR #290, opened 2016-07-22
Make table responsive with table-respon div

svbeon:


dequis:
>Create responsive tables by wrapping any .table in .table-responsive to make them scroll horizontally on small devices (under 768px).

I'm pretty sure horizontal scrolling would suck for a table as big as this one.

ailin-nemui:
thanks for your submission. I tried it and in my opinion the experience is degraded even more with responsible table (admittedly the scripts site is not te best suited for mobile) so I'll reject this

PR #301, opened 2016-08-25
 On branch dictcomplete-debug

Toyoyo:
Changes to be committed:
	modified:   scripts/dictcomplete.pl

Match $word as string, not regex.

This avoids having irssi to use 100% CPU for minutes/hours in the case you hit .[tab] with a very large wordlist (like with 232k entries, as in my french+english dictionary)

### DIFF
--- a/scripts/dictcomplete.pl
+++ b/scripts/dictcomplete.pl
@@ -11,7 +11,7 @@ $VERSION = '1.31';
     description => 'Caching dictionary based tab completion',
     license     => 'Public Domain',
     url         => 'http://juerd.nl/irssi/',
-    changed     => 'Fri Dec 6 11:12 CET 2002',
+    changed     => 'Thu Aug 25 19:29:30 CEST 2016',
     changes     => 'Removed a silly mistake'
 );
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
as far as I understand the script is supposed to allow regex matching. so this change would be bad in general. I wonder what is causing the cpu? is it because the script is returning a huge list? that could be improved by limiting the amount of words pushed to @$complist

Toyoyo:
I tend to disagree.
The whole point of the script is word completion based on a sorted dictionnary.

I still think the $word variable (and not a $regex or something, $word) is in a/^/i regex to allow case insensitivity and make sure it only matches from the beginning of the word, and not anything containing the word.

ailin-nemui:
There is specifically code in the script which judges whether the variable contains regex metacharacters and scans the whole index in that case

ailin-nemui:
It is useful for example if you forget the exact spelling and write ab..?.?cd<tab>

ailin-nemui:
I will close this for now, please reopen if you can address the issues mentioned above

PR #173, opened 2015-11-16
FORCEFULLY ADD ALL THE SCRIPTS FROM NEI'S WEBSITE

dequis:
```
18:33 < dx> ahf: should i forcefully submit to s.i.o all the scripts in nei's website?
18:33 < dx> FORCEFULLY
18:33 < Nei> hehe
18:33 < Nei> could be a good idea
18:34 < ahf> yes, do it with violence and chaos
18:34 < dx> hell yeah
```

:fire::fire::fire::fire::fire::fire::fire::fire::fire::fire:

:boom::boom::boom::boom::boom::boom::boom::boom::boom::boom:

:finnadie::finnadie::finnadie::finnadie::finnadie::finnadie::finnadie::finnadie::finnadie::finnadie:

Ok enough.

Using the `nei-scripts` branch in this same repository, feel free to push commits moving stuff around or deleting scripts that are considered too shitty to be in s.i.o. I just added all of them.

I renamed `mouse.pl` as `mouse-awl.pl`, since it seems to be an extension of it that isn't backwards compatible.

There was a newer version of tmux_away.pl with some changes, I left it out to only add new scripts here.

The relationship between `mouse_soliton.pl` and `mouse.pl` is unclear.

Same thing with `trackbar22.pl` and `trackbar.pl`, although the former seems to have 0.8.17 specific code (but seems based on an older version of trackbar.pl?)

**I will forcefully merge this 24 hours from now, regardless of its state. You have been warned.**

PR #302, opened 2016-08-30
ido_switcher: Allow limited searching for #\d+ channels

avar:
[Heads up to @dequis who wrote this code]

Currently if there's a channel like e.g. #911 you can't search for it
because it goes down the codepath where we try to match window ids.

As a hack around this only go down that codepath if the number the
user enters is <= than our actual number of windows, anything higher
than that can't be matched, so we might as well try to match it as a
channel name instead.

This doesn't fix the case of e.g. trying to match a channel called #10
when you have 10 windows. I think this code should by trying different
approaches in order and falling back if either there's no matches or
too few, or better yet just doing all approaches and sorting the
resultset by a heuristic like if the query matches /^\d+$/.

But this is a minor change that's good enough for my use-case, and I
can't see how it would break anything.

### DIFF
--- a/scripts/ido_switcher.pl
+++ b/scripts/ido_switcher.pl
@@ -778,7 +778,7 @@ sub update_matches {
     }
 
     $regex_valid = 1;
-    if ($search_str =~ m/^\d+$/) {
+    if ($search_str =~ m/^\d+$/ and $search_str <= @window_cache) {
 
         @search_matches =
           grep {


ailin-nemui:
The window cache does not equal the maximum refnum, consider /window number 911

avar:
@ailin-nemui Correct, brainfart on my part. I could `$search =~ /^\d+$/ and exists $refnum_idx{$search}` which I guess would work once I created a `%refnum_idx`, or do the larger patch to make the overall search more elastic like I suggested in the commit.

dequis:
I did not write that code. I just imported them from nei's website, see the commit log of 6dec1974e3ac2e71c26e842887a4b8fe2ec9082c and #173 for more context.

ailin-nemui:
I will close this for now, feel free to reopen once the mentioned concerns are addressed

PR #305, opened 2016-09-11
Improve anames nick prefix sorting

### LINKED ISSUE #304: anames.pl - /anames output isn't necesarily ordered the same as /names
It seems to mostly have to do with ~ and & prefixes.

https://i.imgur.com/Q8ONuv2.png

tslocum:
Fixes #304.


ailin-nemui:
this should fall back to the old code for servers that do not support/make use of get_nick_flags, such as XMPP

tslocum:
I've added a sensible default order which is used when get_nick_flags doesn't work out.

PR #311, opened 2016-09-23
Added iquiz.pl & iquiz_en.pl

wilkowy:
iQuiz is a quiz script featuring five types of games: classic Dizzy, Mieszacz, Familiada, Multi (Familiada without teams) and Pomieszany.

iquiz.pl - original script (Polish)
iquiz_en.pl - English translation

### DIFF
--- a/scripts/iquiz.pl
+++ b/scripts/iquiz.pl
@@ -1,0 +1,1261 @@
+##################################################################
+##    irssi Quiz (iQuiz) script (2010-2016) by wilk/xorandor    ##
+##################################################################
+## Script inspired by classic mIRC scripts: "Dizzy" by Dizzy,   ##
+##   "Mieszacz" & "Familiada" by snajperx (both with my later   ##
+##   upgrades).                                                 ##
+## Other credits:                                               ##
+##   Bjoern 'fuchs' Krombholz for splitlong.pl calculations     ##
+##################################################################
+
+# Tested more or less with irssi 0.8.15 & 0.8.18 and Perl 5.8.8, 5.10.1, 5.14.2, 5.16.3 & 5.18.2
+
+# Script works with:
+#
+# - standard Dizzy/Pomieszany files (also without "pyt"/"odp" prefixes):
+#
+# pyt Evaluate: 2+2=?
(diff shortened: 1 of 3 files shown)


wilkowy:
I see some checks about open/close and using comma failed, but the script is working fine for 5 years now.

ailin-nemui:
hi @wilkowy , thanks for your submission. we have decided to require some best practices for all scripts that go to scripts.irssi.org and I hope you can find the time to implement them. This doesn't mean there is any *fault* with your script per se.

In the questionable case, it should be as simple as changing

    (open(my $fh, $fname))

to

    (open(my $fh, "<", $fname))

in your script. If the green checkmark appears instead of the red cross then we're good to go.

Thanks for your understanding

wilkowy:
Sure, no problem. What about constructs like `print(something), return if (error_condition);`? They also gave check fails. Should I rewrite the postfix?

As I'm not fully familiar with GH yet - I have to close this pull and make a new one, right?

dequis:
>Sure, no problem. What about constructs like `print(something), return if (error_condition);`? They also gave check fails. Should I rewrite the postfix?

The error says "Comma used to separate statements". You could just use semicolons instead of comma.

[More details here](https://metacpan.org/pod/Perl%3A%3ACritic%3A%3APolicy%3A%3AValuesAndExpressions%3A%3AProhibitCommaSeparatedStatements)

>As I'm not fully familiar with GH yet - I have to close this pull and make a new one, right?

No need to, you can just push more commits to the same branch and this PR will update.

wilkowy:
> The error says "Comma used to separate statements". You could just use semicolons instead of comma.

No, that is completely different case. `do(something), return if (error_condition);` is equivalent for `if (error_condirion) { do(something); return; }` The comma is used there on purpose - to execute both commands one by one (not a real list separator). I just made a one-liner shortcut. However if this is confusing for someone or the travis validator than I'll rewrite this as well.

ailin-nemui:
you can also ignore the , issue, you only need to change the "open"

wilkowy:
God damn it I've messed now.

PR #316, opened 2016-10-26
various bugfixes from many months of usage

fullmetal1:


dequis:
It's pretty hard to tell what was changed (if anything) because you changed the mode of all files to be executable. Don't do that.

fullmetal1:
yeah I noticed. Gunna have to revert

PR #323, opened 2016-12-07
New listsort.pl script: order /list by chan size

IsaacG:
New script which sorts the results of /list by channel size.

### DIFF
--- a/scripts/listsort.pl
+++ b/scripts/listsort.pl
@@ -55,6 +55,6 @@ sub list_end {
         Irssi::print($msg, MSGLEVEL_CRAP);
     }
     # Drop the hash values; no point in holding them in memory.
-    delete @list{keys %list};
+    %list = ();
 }
 
--- a/scripts/listsort.pl
+++ b/scripts/listsort.pl
@@ -1,3 +1,5 @@
+use strict;
+use warnings;
 use Irssi;
 use vars qw/$VERSION %IRSSI/;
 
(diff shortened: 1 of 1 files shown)


scripts/listsort.pl
```diff
@@ -0,0 +1,60 @@
...
+            $list{$name}{'size'},
+            $name,
+            $list{$name}{'desc'},
+            $mode
+        );
+
+        Irssi::print($msg, MSGLEVEL_CRAP);
+    }
+    # Drop the hash values; no point in holding them in memory.
+    delete @list{keys %list};
```
ailin-nemui: Use %list=() instead of this inefficient approach 😉

IsaacG:
That's correct. Losing support for parallel /list's is something I think it worth it. In theory, I could track per-network lists and print them independently. However, the behaviour for parallel lists without this script isn't going to be super pretty, either. I'm not sure how servers would handle multiple /list commands on the same server; do they get serialized? If not, there's no solution for that.

scripts/listsort.pl
IsaacG: Done

ailin-nemui:
yes, I meant for multiple servers. I guess it's not so frequently happening in real usage

IsaacG:
Yup. I could handle list entries on a per-server basis but I'm not sure it's worth the additional code and room for bugs. People typically don't request a second list from another server before the first completes.

PR #324, opened 2016-12-07
Clean up my reorder script

IsaacG:
* Add comments
* Add a VERSION
* Drop some silly features, eg autorun on load and network tags that got a trailing '2'
* Convert tabs to 4-space
* Use a valid contact email


rud01f:
$filename .= '.layout' unless ($filename =~ /.layout$/);  # shouldn't the dot in regular expression be escaped? it matches i.e. "my_layout" now (which i think is not intended)

PR #325, opened 2016-12-07
Add ident.pl - automatic identifying

IsaacG:
* Read passwords from ~/.irssi/passwords. Space-separated tag-password pairs, one per line.
* When prompted by nickserv to identify, use the tag and password to do a: /quote nickserv identify $password


IsaacG:
Oops. Got a tiny lil' bug that needs cleaning up.

PR #326, opened 2016-12-07
Add ident.pl - automatically identify from file-based pws

IsaacG:
* Read passwords from ~/.irssi/passwords. One entry per line, space-separated network tag and password
* When prompted by NickServ to identify, do so using /^quote nickserv


ailin-nemui:
a reload function could be nice...

IsaacG:
I thought briefly of a reload ... then realized you could just do:  `/script load ident.pl`. One less command to pollute irssi with ;)

PR #331, opened 2017-01-14
Handle nickserv on reconnect

dedeibel:
Please accept my PR, since it might be of use to others.

The following changes solve the issue of being able to reauthenticate and rejoin a restricted channel after a disconnect on https://hackint.org/services where you have to request an invite to such a channel.

0. When you have a disconnect you will eventually reconnect with a different nick until the old user is removed.
1. keepnick will notice that and change the nick back
2. nickserv will handle authentication
2.1. the new postcmd allows to ask ChanServ for an invite to the restricted channel
3. invitejoin will accept the invite and join the restricted channel
3.1 the new allowed users list will only accept invite from ChanServ ...

I tried to use the already present style of loading / storing settings and extended the help dialogs.

### DIFF
--- a/scripts/invitejoin.pl
+++ b/scripts/invitejoin.pl
@@ -87,38 +87,29 @@ sub load_allowed_nicks {
     my ($file) = @_;
     @allowed_nicks = ();
     if (-e $file) {
-        local *F;
-        open(F, "<", $file);
+        open(my $fh, "<", $file);
         local $/ = "\n";
 
-        while (<F>) {
+        while (<$fh>) {
             chomp;
             my $new_allowed = new_allowed_nick(split("\t"));
             if (($new_allowed->{net} ne "") && ($new_allowed->{nick} ne "")) {
                 push(@allowed_nicks, $new_allowed);
             }
         }
-        close(F);
+        close($fh);
     }
 }
 
 sub save_allowed_nicks {
     my ($file) = @_;
+    open(my $fh, ">", $file) or die "Can't create $file. Reason: $!";
 
-    if (-e $file) {
-        local *F;
-        open(F, ">", $file);
-
-        for my $allowed (@allowed_nicks) {
-            print(F join("\t", $allowed->{net}, $allowed->{nick}) . "\n");
-        }
-    
-        close(F);
-    } else {
-      open(F, ">", $file) or die "Can't create $file. Reason: $!";
-      close(F);
-      save_allowed_nicks($file);
+    for my $allowed (@allowed_nicks) {
+      print($fh join("\t", $allowed->{net}, $allowed->{nick}) . "\n");
     }
+
+    close($fh);
 }
 
 sub new_allowed_nick {
@@ -205,7 +196,9 @@ sub invitejoin_runsub {
 sub is_allowed_nick {
   my ($server, $nick) = @_;
 
-  return 1 unless @allowed_nicks < 1;
+  # If no allowed nicks are specified (initial configuration) accept
+  # all invite requests, which mimics previous behavior of this script
+  return 1 if @allowed_nicks == 0;
 
   return (grep {
     $_->{net}  eq $server->{tag} &&
--- a/scripts/invitejoin.pl
+++ b/scripts/invitejoin.pl
@@ -24,7 +24,7 @@ use strict;
 use Irssi;
 use vars qw($VERSION %IRSSI);
 
-$VERSION = "0.02";
+$VERSION = '0.02';
 
 %IRSSI = (
     authors     => 'Geert Hauwaerts',
@@ -32,8 +32,8 @@ $VERSION = "0.02";
     name        => 'invitejoin.pl',
     description => 'This script will join a channel if somebody invites you to it.',
     license     => 'Public Domain',
-    url         => 'http://irssi.hauwaerts.be/invitejoin.pl',
-    changed     => 'Di 3. Jan 19:46:51 CET 2017',
+    url         => 'https://github.com/irssi/scripts.irssi.org/blob/master/scripts/invitejoin.pl',
+    changed     => 'Di 17. Jan 19:32:45 CET 2017',
 );
 
 my $help = <<EOF;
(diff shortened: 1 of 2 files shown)


ailin-nemui:
I recommend that you use sasl or certfp instead

dedeibel:
Thanks, SASL it is. Nonetheless The code works and is useful if nickserv is used.

scripts/invitejoin.pl
```diff
@@ -33,41 +33,213 @@
...
+    'invite_denied', '%R>>%n %_Invitejoin:%_ Invite from nick %_$1%_ on %_$0%_ to %_$2%_ not followed because it is not in the allowed list.',
 ]);
 
-sub invitejoin {
+sub load_allowed_nicks {
+    my ($file) = @_;
+    @allowed_nicks = ();
+    if (-e $file) {
+        local *F;
+        open(F, "<", $file);
```
ailin-nemui: while you're in the process of modifying the script, why not replace this with "modern style" `open my $f, "<", $file`

scripts/invitejoin.pl
```diff
@@ -33,41 +33,213 @@
...
+        close(F);
+    }
+}
+
+sub save_allowed_nicks {
+    my ($file) = @_;
+
+    if (-e $file) {
+        local *F;
+        open(F, ">", $file);
```
ailin-nemui: this function is weird, why create the file and call yourself recursively instead of just saving it?

scripts/invitejoin.pl
```diff
@@ -33,41 +33,213 @@
...
+        Irssi::command_runsub('invitejoin', $data, $server, $item);
+    } else {
+        Irssi::printformat(MSGLEVEL_CLIENTCRAP, 'invitejoin_usage');
+    }
+}
+
+sub is_allowed_nick {
+  my ($server, $nick) = @_;
+
+  return 1 unless @allowed_nicks < 1;
```
ailin-nemui: so this will always return true if there are *any* allowed nicks, but always fail if there are no allowed nicks? or how to understand the code?

scripts/invitejoin.pl
dedeibel: borrowed it from the nickserv script actually ...

scripts/nickserv.pl
```diff
@@ -442,13 +555,25 @@ sub nickserv_notice {
         } elsif ($text =~ /^Password accepted - you are now recognized/ || $text =~ /^You are now identified for/) {
             Irssi::signal_stop();
             Irssi::printformat(MSGLEVEL_CLIENTCRAP, 'password_accepted', $server->{tag});
+	    run_postcmds($server, $server->{tag}, $server->{nick})
```
ailin-nemui: nitpick, the indentation seems off

scripts/invitejoin.pl
dedeibel: borrowed it from the nickserv script actually ...

I could make it nicer if you say there is a chance of merging.

scripts/invitejoin.pl
dedeibel: Not quite.

It keeps the previous behavior which allows all invites without restriction. If you have set any allowed nicks, it restricts the invite joins to the allowed nicks.

> so this will always return true if there are**n't** any allowed nicks, but always fail if there are no **matching** allowed nicks?

scripts/invitejoin.pl
ailin-nemui: what makes  you think it has no chance of being accepted?

scripts/invitejoin.pl
ailin-nemui: I wonder if that is really the case...

scripts/invitejoin.pl
dedeibel: Just being unsure about how this repo is maintained and who considers what feature useful.

scripts/nickserv.pl
```diff
@@ -545,18 +693,27 @@ sub nickserv_runsub {
...
+Irssi::command_bind('ns listpostcmd', 'list_postcmd');
+
 Irssi::command_bind('nickserv delnet', 'del_network');
 Irssi::command_bind('ns delnet', 'del_network');
 
 Irssi::command_bind('nickserv delnick', 'del_nickname');
 Irssi::command_bind('ns delnick', 'del_nickname');
 
+Irssi::command_bind('nickserv delpostcmd', 'del_postcmd');
+Irssi::command_bind('ns delpostcmd', 'del_postcmd');
```
ailin-nemui: I can't say I like all those command_binds very much... alas they were there already

scripts/invitejoin.pl
ailin-nemui: `unless @allowed_nicks < 1` that means `IF @allowed_nicks >= 1 ` then it should (unconditionally) return true, or am I mistaken? please double check this

dedeibel: unless got me again ... when you thought you could handle it

Fixing it. My testcase must have been bad too.

dedeibel:
Thanks for looking into it. I'll come up with some improvements - need to do some tests.

scripts/nickserv.pl
ailin-nemui: maybe I would remove the binds for /ns nevertheless. any user could manually /alias them. another reservation I have about this script, is that /nickserv is often /alias msg nickserv by some people. but then again those don't have to install the script

dedeibel:
Fixed the bug you metioned, did some refactorings, added some missing checks, did some more testing.
The ns command is still in there, didn't want to change previously working behavior.

ailin-nemui:
by the way did you know that irssi has an orthogonal feature of autojoin on invite to auto channels?

dedeibel:
oh ... well, thanks. Must have overlooked it.

ailin-nemui:
(it looks like it has been broken to start with / not caused by your changes at least, so I wonder since you changed the script, could you even use it?)

dedeibel:
about the pager.pl ... I did not really use it. Just looked into it, saw it's formatting was totally mad, ran vim autoindent and left it there. Maybe I should have left it untouched I guess.

ailin-nemui:
yea.. in that case please remove it from your commit

dedeibel:
it's gone, `git pull -f` in case you had a checkout of the branch

dedeibel:
Wohoo, thx

Issue #334, opened 2017-01-23
desktop-notify.pl: unable to display Chinese


jimgrape:
![2017-01-23 11-19-49](https://cloud.githubusercontent.com/assets/22955609/22192734/6b935a20-e171-11e6-89e6-5b04ef47fd64.png)

jimgrape:
by xhe

--- 1	2017-01-23 11:19:34.096562817 +0800
+++ desktop-notify.pl	2017-01-23 11:18:12.092687865 +0800
@@ -24,6 +24,7 @@
 use Irssi;
 use HTML::Entities;
 use Glib::Object::Introspection; # Ignore 'late INIT' warning message if autoloading
+use Encode;
 
 our $VERSION = '1.0.0';
 our %IRSSI = (
@@ -38,6 +39,7 @@
 # List of standard icons can be found here:
 # http://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#names
 my $notify_icon;
+my $term_charset;
 
 my $help = '
 /set notify_icon <icon-name>
@@ -60,6 +62,7 @@
 
 sub setup_changed {
 	$notify_icon = Irssi::settings_get_str('notify_icon');
+	$term_charset = Irssi::settings_get_str('term_charset');
 }
 
 sub priv_msg {
@@ -71,9 +74,9 @@
 		return;
 	}
 
-	my $msg = HTML::Entities::encode_entities(Irssi::strip_codes($msg));
+	my $msg = HTML::Entities::encode_entities(Irssi::strip_codes($msg), "\<>&'");
 	my $network = $server->{tag};
-	my $noti = Notify::Notification->new($nick . '@' . $network, $msg, $notify_icon);
+	my $noti = Notify::Notification->new($nick . '@' . $network, decode($term_charset, $msg), $notify_icon);
 	$noti->show();
 }
 
@@ -94,8 +97,8 @@
 	}
 
 	my $network = $server->{tag};
-	my $msg = HTML::Entities::encode_entities($stripped);
-	my $noti = Notify::Notification->new($dest->{target} . '@' . $network, $msg, $notify_icon);
+	my $msg = HTML::Entities::encode_entities($stripped, "\'<>&");
+	my $noti = Notify::Notification->new($dest->{target} . '@' . $network, decode($term_charset, $msg), $notify_icon);
 	$noti->show();
 }

xhebox:
let me clear:
[...]

PR #335, opened 2017-01-23
desktop-notify.pl: process non-ascii msgs

xhebox:
irssi-1.0.0
bug found by jimgrape,  patched by xhebox.
https://github.com/irssi/scripts.irssi.org/issues/334


scripts/desktop-notify.pl
```diff
@@ -24,6 +24,7 @@
 use Irssi;
 use HTML::Entities;
 use Glib::Object::Introspection; # Ignore 'late INIT' warning message if autoloading
+use Encode;
 
 our $VERSION = '1.0.0';
```
ailin-nemui: please change to 1.0.1 here

PR #336, opened 2017-01-25
Rename chanact.pl /[chanact_]window_(un)alias

ear:
The plugin adds the `/window_alias` and `/window_unalias` commands.

This introduces ambiguity for shorter versions of the `/window` command. In particular it has the effect of disabling the shorthand `/win` (which is a popular way to call `/window` documented all over the web.)

My reasoning to change the two commands to `/chanact_window_alias` and `/chanact_window_unalias` is four-fold:
* `/win` is a useful and universally known and used shorthand for `/window` and it is used very often
* window aliases for `chanact.pl` are rarely set and unset
* the `/chanact_` prefix enables tab-completion with `/chana[TAB]` of both of them (as well as eventual new ones)
* all of the `chanact` options exist under `/set chanact_` (which is mentioned in the documentation of the plugin.)

Thank you for taking the time to review this small proposal.

### DIFF
--- a/scripts/chanact.pl
+++ b/scripts/chanact.pl
@@ -5,7 +5,7 @@ use Irssi::TextUI;
 
 use vars qw($VERSION %IRSSI);
 
-$VERSION = "0.5.15";
+$VERSION = "0.5.16";
 %IRSSI = (
     authors     => 'BC-bd, Veli',
     contact     => 'bd@bc-bd.org, veli@piipiip.net',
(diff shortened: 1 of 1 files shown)


ailin-nemui:
- can you try to reach the original author about this please
- this may break other people habits
- another workaround for users of this script would be `/alias win window`
- you didn't actually update the VERSION

ear:
Thanks for the fast response and the /alias idea!

I can try to send an email to the two listed contacts in line 11. I didn't try too hard yet because the svn repo linked in line 15 is dead.

My only counterpoint to the "habits" is that one does not really run those commands very often: usually once when you're setting up `chanact.pl`. At least that is my own experience.

I do think that the default chose here is too invasive, ruining the behavior of /win commands, and requiring manual intervention.

VERSION is fixed now.

mafgh:
I like it, but I'd rather release this as a 0.6.0 as it changes command names.

I'll come back to it once I figured out how to handle #342.

PR #337, opened 2017-02-01
[go.pl] Make matching behaviour configurable

madduck:
This introduces a few settings, allowing the behaviour of `go.pl` to be configured:

- `go_match_case_sensitive` (default:true), match channel names case-sensitively
- `go_complete_case_sensitive` (default:false), complete channel names case-sensitively
- `go_match_anchored` (default:true), match channel names only at start of word
- `go_complete_anchored` (default:true), complete channel names anchored at start of word


ailin-nemui:
I wonder if complete_anchored should be off to better match the old behaviour?

madduck:
@ailin-nemui you are absolutely right. Wonder how I missed that. PR/branch updated.

PR #343, opened 2017-02-08
[adv_windowlist] Make the statusbar leader text configurable

madduck:
This introduces the settings `awl_sb_leader`, which defaults to the current "Awl: " text and can be henceforth overridden to configure (or hide) the status bar leader text.

### DIFF
--- a/scripts/adv_windowlist.pl
+++ b/scripts/adv_windowlist.pl
@@ -1,7 +1,7 @@
 use strict;
 use warnings;
 
-our $VERSION = '1.3'; # 463402cffae35e5
+our $VERSION = '1.3.1'; # 463402cffae35e5
 our %IRSSI = (
     authors     => 'Nei',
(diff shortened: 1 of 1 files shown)


madduck:
This isn't complete:

> < Nei[m]> madduck: there is somewhere code that calculates mouse coordinates
> < Nei[m]> You'd have to account for the configurable setting there as well

Can't turn to that now though…

madduck:
I've stared at this for a bit now, but I can't figure it out. This is black voodoo magic. I hope @ailin-nemui can provide some guidance.

madduck:
@ailin-nemui Is this was `awl_shared_sbar` is all about? I.e. I just need to add to the right-hand side the length of the header?

ailin-nemui:
closing this and moving it to https://github.com/ailin-nemui/irssi-adv_windowlist/issues/1

PR #344, opened 2017-02-09
[trigger.pl] Save triggers idempotently

madduck:
`/trigger save` writes out switches and parameters in the (random) order as return by iterating a Perl hash. In addition, there's a trailing new line on each saved trigger. Both of these make it slightly harder to track the Irssi configuration in a VCS such as Git.

This PR thus makes two changes that shouldn't have any effect on functionality:

1. Remove the trailing newline
1. Sort the switches and parameters before writing them to a file so that their order becomes well-defined.

I also accidentally removed trailing whitespace in the code (2 occurrences), but I think this is okay to be included and thus I didn't go through the hassle to get that part of the commit removed.

### DIFF
--- a/scripts/trigger.pl
+++ b/scripts/trigger.pl
@@ -982,14 +982,14 @@ sub to_string {
 	my ($trigger, $compat) = @_;
 	my $string;
 
-	foreach my $switch (@trigger_switches) {
+	foreach my $switch (sort @trigger_switches) {
 		if ($trigger->{$switch}) {
 			$string .= '-'.$switch.' ';
 		}
 	}
 
 	if ($compat) {
-		foreach my $filter (keys(%filters)) {
+		foreach my $filter (sort keys(%filters)) {
 			if ($trigger->{$filter}) {
 				$string .= '-' . $filter . param_to_string($trigger->{$filter});
 			}
@@ -1000,7 +1000,7 @@ sub to_string {
 		}
 	}
 
-	foreach my $param (@trigger_params) {
+	foreach my $param (sort @trigger_params) {
 		if ($trigger->{$param} || ($param eq 'replace' && defined($trigger->{'replace'}))) {
 			$string .= '-' . $param . param_to_string($trigger->{$param});
 		}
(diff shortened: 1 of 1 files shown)


ailin-nemui:
maybe you forgot pt.2?

madduck:
@ailin-nemui Ouch. You're right. Thanks for the catch. I'll send an update when I get this figured out. I am terribly sorry…!

madduck:
@ailin-nemui done. Thanks again!

scripts/trigger.pl
```diff
@@ -981,15 +981,15 @@ sub param_to_string {
 sub to_string {
 	my ($trigger, $compat) = @_;
 	my $string;
-	
-	foreach my $switch (@trigger_switches) {
+
+	foreach my $switch (sort @trigger_switches) {
```
ailin-nemui: these are inherently ordered and don't need to be sorted

scripts/trigger.pl
```diff
@@ -981,15 +981,15 @@ sub param_to_string {
...
 		}
 	}
-	
+
 	if ($compat) {
-		foreach my $filter (keys(%filters)) {
+		foreach my $filter (sort keys(%filters)) {
 			if ($trigger->{$filter}) {
 				$string .= '-' . $filter . param_to_string($trigger->{$filter});
 			}
```
ailin-nemui: the trigfilter which would require ordering are not sorted

scripts/trigger.pl
```diff
@@ -1000,11 +1000,12 @@ sub to_string {
 		}
 	}
 
-	foreach my $param (@trigger_params) {
+	foreach my $param (sort @trigger_params) {
```
ailin-nemui: these are also already ordered

madduck:
Hm, I am a bit at a loss now, for obviously arrays are sorted. I have two machines and I keep my triggers in Git. The reason I came up with this patch was that every time I would pull commits from one of the machines, and load the triggers on the other, the next `/save` would result in a diff…

After appying my patch, this has never occurred again. So yes, I might have overshot e.g. in trying to sort inherently sorted arrays, but some things do need an explicit sort.

I'll check this again when I find the time…

ailin-nemui:
closing for inactivity, please reopen once you are ready

PR #347, opened 2017-02-10
[adv_windowlist] Enable matching the F-keys as window changing shortcuts

madduck:
This enhances the AWL keymap parser to also match the use of F-keys
(i.e. F1 through F12), which will have to be mapped from their escape
sequences first (e.g. /bind ^[[11~ key F1).

The convention is followed that "meta-F1" gets displayed as "F1", and
"^F1" is shown as "^F1". Since it also makes sense to use F-keys without
modifiers, the syntax "`F1" is used to identify a "pure" F-key.


ailin-nemui:
moving to https://github.com/ailin-nemui/irssi-adv_windowlist/issues/3

PR #353, opened 2017-02-21
[ctrlact] Show the window refnum in debug output if there's not window name

madduck:


ailin-nemui:
if you want to publish new versions you need to increase VERSION, otherwise I recommend you do development in your own github project and submit scripts once when you believe a version to be ready

madduck:
Will do. I'll include this change in the next version push.

PR #362, opened 2017-03-07
[colorkick] declare $VERSION %IRSSI

bw1:


ailin-nemui:
this script contains another bug:
 `$target == $channel`, will always succeed. maybe you can fix this script properly otherwise we should better not even fix the compilation

bw1:
I'm sorry, I overlooked that.

PR #374, opened 2017-03-18
[ixmmsa] add module Xmms

bw1:


ailin-nemui:
I think xmms is dead

bw1:
not really. 
here on the desktop is a xmms. 

the library is on cpan and the script works.

PR #376, opened 2017-03-19
[wordcompletition] init for the database

bw1:


ailin-nemui:
it would be better to rewrite the SQL using placeholder syntax to avoid sql injection

PR #386, opened 2017-04-01
[kicks][crit] use strict; open

bw1:


scripts/kicks.pl
```diff
@@ -142,39 +145,41 @@ sub get_a_reason {
 }
 
 sub cmd_realkick {
-    my ($data, $server, $channel, $cmd) = @_;
+    (my $data, my $server, my $channel, my $cmd) = @_;
```
ailin-nemui: the original one was correct

scripts/kicks.pl
```diff
@@ -83,51 +84,53 @@ sub initialize {
...
+
+        my $line=0;
+        
+        while (<CONF>) {
+            $line++;
+
+            next if /^\s*#/;
+
+            chomp;
+            (my $key, my $reasons) = /^(\S+)\s+(.+)\s*$/ or next;
```
ailin-nemui: it's more common to read `my ($x, $y)`

PR #395, opened 2017-04-07
add sbitems metadata

ailin-nemui:


ailin-nemui:
I am planning to merge this without version bumps since it is only metadata updates for scriptassist

PR #396, opened 2017-04-09
trigger.pl: default address to empty in mask match

dwfreed:
Server notices are included in the privnotices type, but their address parameter is undefined, because the prefix of the message is just the server name.  By defaulting address to an empty string in the masks match, we can match server notices easily with `-privnotices -masks '<servername>!'` instead of a really messy rawin + regexp.  (Irssi merges the nick and address passed into masks_match with a `%s!%s` format string, so there'll be a trailing "!" when address is an empty string.)


ailin-nemui:
need to update VERSION

dwfreed:
Bah, I was holding off bumping the version because I had other changes I wanted to have merged too, but I hadn't gotten around to making sure they were clean and committing them.  I should have said something, sorry.

PR #404, opened 2017-05-02
[go2.pl] clear prompt before changing windows

madduck:
This switches the order of changing windows and clear the prompt. Now
the prompt line is cleared before the `go` command is called, which
means that scripts like per_window_prompt.pl won't save the `/go`
command.


ailin-nemui:
please follow the versioning guide on the main site: 1.0 -> 1.1

madduck:
Done, sorry about that. Different projects have different guidelines and it's too easy to get confused…

PR #405, opened 2017-05-08
Clear execute perms

madduck:
Irssi scripts really don't need the execute bit set, nor shebang lines. Cleaning these up…


ailin-nemui:
while you're right that it is not needed, many scripts have the shebang, some editors also use it to ensure it's a Perl program. It seems odd that you removed it only from 1 script. I would rather defer any such clean up actions when the scripts itself have improvements

madduck:
Simple explanation: the commit was incomplete. Sorry. But I guess I'll retire this branch…

PR #413, opened 2017-06-06
stop perl being confused by ][ in regex

ailin-nemui:
reported by SCHAPiE, fix by vague and Grinnz


ailin-nemui:
https://rt.perl.org/Public/Bug/Display.html?id=127581 

issue again in 5.26

PR #418, opened 2017-07-05
update tmux-nicklist-portable to 0.1.6

ailin-nemui:
- improve performance especially on /wc


Aebian:
can confirm, works n1

PR #419, opened 2017-07-20
nicklist.pl 0.4.9 - fix scrolling

ailin-nemui:
unbreak a change that broke scrolling in nicklist 0.4.8

reported by @kiero


ailin-nemui:
@mrjink for your info

PR #420, opened 2017-07-26
Add identd capability to irssi during connections

vague666:


dequis:
>First-time contributor

Welcome! Always good to see new faces. Thank you for your contribution!

PR #421, opened 2017-08-01
trigger.pl: various improvements

dwfreed:
* Allow notify trigger types to use -masks

They have masks, so allow matching them.

* Allow matching old nick using -other_masks for nick changes

Matching the old nick has its uses, so shoehorn it under -other_masks.

* Add -last option; if trigger with this option matched, doesn't
  process any more triggers for that message

Useful if the trigger's action does something like kickban or akill.

* Allow moving triggers entirely by name

Useful for programmatic rearranging of triggers, as you don't need to
know the number of the target.  First trigger will be placed before
second trigger.  (Goes well with -last, too)

### DIFF
--- a/scripts/trigger.pl
+++ b/scripts/trigger.pl
@@ -23,7 +23,7 @@ use Text::ParseWords;
 use IO::File;
 use vars qw($VERSION %IRSSI);
 
-$VERSION = '1.1';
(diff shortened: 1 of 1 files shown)


ailin-nemui:
please add it to the built in documentation

PR #433, opened 2017-10-30
Update screen_away.pl

qoreQyaS:
increased version so the (at least) last change gets detected by scriptassist


ailin-nemui:
thanks

PR #434, opened 2017-10-31
close socket on access from unverified sources, add /help identd

vague666:


ailin-nemui:
up VERSION on changes

PR #439, opened 2017-11-08
Fixed the ipupdate script to use different server

EchterAgo:
Fixed to use ifconfig.co/ip instead of xergio.net because the server does not work anymore.


ailin-nemui:
need to update VERSION

EchterAgo:
Updated

PR #441, opened 2017-11-17
Update bitlbee_typing_notice.pl

### LINKED ISSUE #435: [bitlbee_typing_notice.pl] Crash when connecting to standard IRC server
Whenever I connect to a non-bitlbee IRC server I get this in the main status window and I see the CTCP TYPING messages directly in the private chat windows:
> -!- Irssi: Error in script bitlbee_typing_notice:
> Can't use an undefined value as a HASH reference at /home/mike/.irssi/scripts/autorun/bitlbee_typing_notice.pl line 215.

Simply reloading the script fixes it until next time, but it's rather annoying as it means I can't actually have typing_notice.pl autorun because it always crashes when connecting to the first IRC server.

ailin-nemui:
fix crash due to no server

Fixes #435

### DIFF
--- a/scripts/bitlbee_typing_notice.pl
+++ b/scripts/bitlbee_typing_notice.pl
@@ -80,7 +80,7 @@ use Data::Dumper;
 
 use vars qw($VERSION %IRSSI);
 
-$VERSION = '1.7.2';
+$VERSION = '1.7.3';
 %IRSSI = (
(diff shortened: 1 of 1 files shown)


mijofa:
Tested, it continued to work after joining a couple of new servers, so seems good to me. Thanks for that. :+1:

PR #442, opened 2017-11-17
Update sb_search.pl

### LINKED ISSUE #437: [sb_search.pl] crash with empty windows
| | |
| --- | --- |
| cybrNaut | bug?  -!- Irssi: Error in script sb_search: Can't call method "next" on an undefined value at /home/user/.irssi/scripts/sb_search.pl line 134. |

ailin-nemui:
fix crash on empty windows

Fixes #437

### DIFF
--- a/scripts/sb_search.pl
+++ b/scripts/sb_search.pl
@@ -22,7 +22,7 @@ use Irssi;
 use Irssi::TextUI;
 use vars qw($VERSION %IRSSI);
 
-$VERSION = '1.1';
+$VERSION = '1.2';
 %IRSSI = (
 	authors     => 'Wouter Coekaerts, Emanuele Giaquinta',
 	contact     => 'wouter@coekaerts.be, exg@irssi.org',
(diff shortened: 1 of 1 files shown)


ailin-nemui:
cybrNaut> Nei: seems to work unless I start in an  empty window. if I run it from an empty window, then i get "Can't call method "prev" on an undefined value at sb_search.pl line 136."

 then> yes, I got the same error, sb_search doesn't  like empty windows

cybrNaut> but it looks like he fixed the more normal case.. where you start from a normal  populated window, and search all windows, some of which are empty, it's only if I start the search from an    empty window that version 1.2 fails

 vague> cybrNaut, try making line 136 '$line =  $forward ? $view->{startline}->next : $view->{startline}->prev if $view->{startlin

ailin-nemui:
@vague666 want to test my latest fixes? I moved your suggestion into the else (-> elsif), I hope it works now =)

PR #444, opened 2017-11-23
anames.pl: Drop re-registration of WHO.

unit193:
* Since unnecessary and incomplete, breaking WHO
  generally and for other scripts.
* Change default away nick colour back to grey.

### DIFF
--- a/scripts/anames.pl
+++ b/scripts/anames.pl
@@ -7,11 +7,11 @@
 
 # Options
 # =======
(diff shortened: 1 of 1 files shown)


ailin-nemui:
1. can you please make only the changes/fixes you want, without changing the other styles/order of the script?
2. no need to change the default colours back and forth, you can easily format it for your own desire

PR #451, opened 2017-12-23
toilet.pl: Added filters

bcattaneo:
Added general filters instead of just the "gay" filter. Each filter is separated with a ":" character.


bcattaneo:
Please ignore the "Dec 22, 2017" commit, it was a mistake.

PR #455, opened 2018-01-04
Add prefix_same_nick.pl and nickcolor's fork

bcattaneo:
This is for a specific issue on irssi's repo: https://github.com/irssi/irssi/issues/800
I thought it's a great idea, but I also noticed it'd "break" nickcolor.pl, that's why this commit also includes a new fork ("nickcolor_with_prefix.pl") with this new functionality.

### DIFF
--- a/scripts/deadbeef.pl
+++ b/scripts/deadbeef.pl
@@ -1,0 +1,102 @@
+#
+# 2017-12-30 bcattaneo:
+#  - initial release
+#
+
+use Irssi;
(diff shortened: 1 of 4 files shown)


ailin-nemui:
hi, can you make multiple branches and exclude deadbeef from this one?
Also I have an idea, instead of forking nickcolor maybe you can add it as an option (that can be turned on, default off) to the nickcolor script?

PR #457, opened 2018-01-05
Update nickcolor.pl

ailin-nemui:
fixes an odd crash when $server unexpectedly becomes invalid during command execution.
cf irssi/irssi#796


ailin-nemui:
included in #459

Issue #438, opened 2017-11-06
nickcolor.pl - "/color list" output wrong if identd starts with a number.


RealKindOne:
```./configure --prefix=/usr --enable-silent-rules --enable-true-color --with-socks --with-bot --with-proxy```
...
```Building with true color support.. : yes```

```
user@linux ~/.irssi $ cat saved_colors
ident@script/does/not/like/idents/starting/with/numbers:59
1000@foobar.com:9
```

![screenshot_2017-11-06_06-48-59](https://user-images.githubusercontent.com/20209685/32439955-11c07ee4-c2bf-11e7-9e0e-578bfa87b76d.png)

It should show "1000@foobar.com" in green.

ailin-nemui:
probably needs to use a printf "%02d" or so

bcattaneo:
Hello. #454 fixes it

bcattaneo:
@RealKindOne Hello!
Could you please check if new nickcolor.pl version fixed that for you?
Thanks

RealKindOne:
Looks like its fixed. Thanks.

PR #459, opened 2018-01-09
[nickcolor.pl] Fix wrong output. Add two new (optional) functionalities

bcattaneo:
- Fix issue #438
- Add two new functionalities (requested in https://github.com/irssi/irssi/issues/800).


ailin-nemui:
I think you need to add 1 more step, instead of saving $target only, my $tagtarget = "$server->{tag}/$target"; otherwise it could be a conflict when you have 2 channels on different networks

bcattaneo:
You're absolutely right, thanks

PR #475, opened 2018-02-01
Merge trackbar22 into trackbar script

ailin-nemui:


ailin-nemui:
happened to incur some additional work

dequis:
Thank youuuu! Not the most backwards compat update but it needed to be done and it's unlikely there are much better ways. I had no idea `/tb` was a command or that it had settings, and i bet most people didn't, either.

The old version had a note about buf.pl support being fiddly, does the new priority stuff fix that?

ailin-nemui:
> The old version had a note about buf.pl support being fiddly, does the new priority stuff fix that?

not sure but I think it does

dx wrote
> hmmm the newer trackbar seems to move something on the screen right after i switch windows
> https://dump.dequis.org/D1zb-.webm
> yeah the old trackbar doesn't do that

I can't reproduce this but it's caused by f0ebb08dc2bef7b1a8ccc31474c9af23f85541d8, apparently. would be good if someone can try out some things, e.g. remove

https://github.com/ailin-nemui/scripts.irssi.org/blob/7c9355afc1222206681a2086e5e8fae9052c7bd2/scripts/trackbar.pl#L434

see if that helps?

otherwise may be better to revert that commit and live with "broken" trackbars in horizontal splits

bw1:
@mjholtkamp

dequis:
>would be good if someone can try out some things, e.g. remove
>
>https://github.com/ailin-nemui/scripts.irssi.org/blob/7c9355afc1222206681a2086e5e8fae9052c7bd2/scripts/trackbar.pl#L434
>
>see if that helps?

Sorry went afk earlier, tested it now, removing the first redraw seems to help. Nothing odd about the single-window use case as far as i can see.

Sideways splits also seem to work, but I can see similar bump-like behavior as soon as i start using them - not with empty lines at the bottom that vanish after a redraw, but with missing lines in the middle (often random chat lines, sometimes the trackbar), where the bump effect only happens with lines before that one, not lines after that one 

More graphically, given these lines:

```
2
3
4
5
6
------
```

Sometimes there's a short flash of:

```
1
2
3
5
6
------
```

Again, only when using sideways splits, so, not a regression... not a problem? I'm okay with this, either way.

bw1:
### step by step screen dump
* slow motion [v1.1-g243ae4be](https://imgur.com/fSsco02)
* v1.0.6  v1.0.6

ailin-nemui:
I removed the duplicate redraw, did you notice any problems that weren't there before @bw1 ?

bw1:
screen dump -> [video](https://imgur.com/gndFl69)

best view -> [gist](https://gist.github.com/fc98f9cb1e6dd8d9728d97705f36af5f) 
(record with $script filename ; play with replay.pl)

i have run the script with
* irssi v0.8.17 +  ISO-8859-1 + raspbian
* irssi v0.8.17 +  utf8 + raspbian

		/trackbare markall
  		Can't locate package Irssi::Nick for @Irssi::Irc::Nick::ISA at 
		/home/pi/.irssi/scripts/trackbar.pl line 498.

ailin-nemui:
@bw1 I repro + fix 0.8.15, was caused by my change for horiz. splits.

ailin-nemui:
also fixed Irssi::Nick

ailin-nemui:
@bw1 do you think redraw is issue caused by 

>     redraw_trackbars($newwindow) 

? does it work better with the redraw removed? (I would obviously prefer to keep it because otherwise horizontal splits are wrong trackbar width)

ailin-nemui:
I cannot repro "irssi v0.8.17 + utf8 + raspbian jessie img" or did you set it to `ä-` on purpose? but I fixed iso encoding. thanks for the thorough check

bw1:
markall mark only the visible windows

help:

	MARKALL:     Like mark for all windows

#### encoding, nick and redraw looks good

ailin-nemui:
are you sure? it works for me. but trackbar won't be drawn in empty windows

ailin-nemui:
>   Can't locate object method "view" via package "Irssi::Irc::Server" 
>  at /home/pi/.irssi/scripts/trackbar.pl line 427.

good catch, looks like I broke that command @_@

bw1:
* r48  format

		#    Setting:     trackbar_require_seen
		#    Description: Only clear the trackbar if it has been scrolled to.
		#
		#     /mark is a command that will redraw the line at the bottom.
		#
		#    Command:     /trackbar, /trackbar goto
		#    Description: Jump to where the trackbar is, to pick up reading

dequis:
Uh, you sure everything was addressed? Or was that a misclick?

ailin-nemui:
guess I made a mistake here but since the last criticism are only in comments let's save it for next time?

PR #480, opened 2018-02-15
Unbreak hlscroll on recent versions of irssi

refacto:
With newer versions of irssi, using hlscroll results in this error message:
```
Can't locate object method "view" via package "Irssi::UI::Window"`
```

Thanks to: Nei[e] on #irssi (freenode)


ailin-nemui:
hi @refacto thanks for the contribution. could you please update the version number?

refacto:
Done.

Issue #306, opened 2016-09-12
[urlinfo.pl] post to channel?


nem8:
Any way to have the script post the parsed url info to the channel in say, preferably with blacklist for channels you dont want it to trigger?

ailin-nemui:
@dgl

bw1:
@nem8

ailin-nemui:
bw1 implemented this in new version of script, /set urlinfo_send_channels #yourchannel

PR #483, opened 2018-02-21
[urlinfo][#306] add post to channel

bw1:


ailin-nemui:
I think the code would be more clear if you changed `if (send2channel` to `if (not send2channel` / unless, and reverted the return values in send2channel

bw1:
ok.is changed.

ailin-nemui:
@dgl

PR #484, opened 2018-02-22
[scriptassist] improve autorun creation

ailin-nemui:
@dwfreed can you test this?

### DIFF
--- a/scripts/scriptassist.pl
+++ b/scripts/scriptassist.pl
@@ -5,7 +5,7 @@
 
 use strict;
 
-our $VERSION = '2003020805';
+our $VERSION = '2003020806';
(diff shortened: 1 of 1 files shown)


dwfreed:
It sort of helps, except now it'll just say it isn't a correct link.  It shouldn't care where the autorun symlink points to when it's removing it.

ailin-nemui:
yea but this was the way the original code intended it. some people also put real files in autorun so in this way at least no files are lost and autorun are only deleted when they look identical to what scriptassist would have created

ailin-nemui:
I will check this in for now but I agree the code could be further refined, eg. check if it's a symlink and remove it in any case

PR #490, opened 2018-03-21
More fitting name

vague666:
Lets do it like this for now :)


dequis:
what are you doing vague, stahp

dequis:
Oh nevermind i misread the other PR and thought you renamed one way and then back to the original way

vague666:
:P I am guessing git mv doesn't work here? It could remove the old script straight away, it's fairly new

PR #492, opened 2018-03-26
Write lines to their original levels so that /window hidelevel can be…

vague666:
… used on the pre-upgrade buffer

### DIFF
--- a/scripts/buf.pl
+++ b/scripts/buf.pl
@@ -94,7 +94,7 @@ sub restore {
         }
         my $sep = settings_get_str 'upgrade_separator';
         $sep .= "\n" if $sep ne '';
-        $window->gui_printtext_after($window->last_line_insert, MSGLEVEL_CLIENTNOTICE, "\cO$sep\n");
+        $window->gui_printtext_after($window->last_line_insert, MSGLEVEL_CLIENTNOTICE, "\cO$sep");
         $view->redraw();
     }
     active_win->command('^window scroll on');
--- a/scripts/buf.pl
+++ b/scripts/buf.pl
@@ -93,7 +93,7 @@ sub restore {
             $window->gui_printtext_after($window->last_line_insert, $level, "$data\n");
         }
         my $sep = settings_get_str 'upgrade_separator';
-        $window->gui_printtext_after($window->last_line_insert, MSGLEVEL_CLIENTNOTICE, "\cO$sep\n") if defined $sep;
+        $window->gui_printtext_after($window->last_line_insert, MSGLEVEL_CLIENTNOTICE, "\cO$sep\n") if $sep ne '';
         $view->redraw();
     }
     active_win->command('^window scroll on');
--- a/scripts/buf.pl
+++ b/scripts/buf.pl
@@ -88,14 +88,14 @@ sub restore {
         my $view = $window->view;
         $view->remove_all_lines();
         $view->redraw();
-        my $sep = settings_get_str 'upgrade_separator';
-        $sep .= "\n" if $sep ne '';
-        $window->gui_printtext_after(undef, MSGLEVEL_CLIENTNOTICE, "\cO$sep\n");
         for my $line (@lines) {
             my $level = $line->{level};
             my $data  = $line->{data};
             $window->gui_printtext_after($window->last_line_insert, $level, "$data\n");
         }
+        my $sep = settings_get_str 'upgrade_separator';
+        $sep .= "\n" if $sep ne '';
+        $window->gui_printtext_after($window->last_line_insert, MSGLEVEL_CLIENTNOTICE, "\cO$sep\n");
         $view->redraw();
     }
     active_win->command('^window scroll on');
--- a/scripts/buf.pl
+++ b/scripts/buf.pl
@@ -2,7 +2,6 @@ use strict;
 use vars qw($VERSION %IRSSI);
 use Storable;
 use 5.014000;
-use Data::Dumper;
 
 use Irssi qw(command signal_add signal_add_first active_win
              settings_get_str settings_get_bool channels windows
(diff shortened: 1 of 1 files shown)


ailin-nemui:
we need to change this script in such a way that it still works fine on old irssis...

vague666:
I haven't tested the script without #862 applied, I don't know how critical the patch is

ailin-nemui:
forget what I said but please don't change indentation (4 spaces) of existing scripts ... :-) I will add more review comments once you can do that

scripts/buf.pl
```diff
@@ -40,64 +43,59 @@
 sub _filename { sprintf '%s/scrollbuffer', get_irssi_dir }
 
 sub upgrade {
-    my $fn = _filename;
-    my $old_umask = umask 0077;
-    open my $fh, q{>}, $fn or die "open $fn: $!";
-    umask $old_umask;
-
-    print $fh join("\0", map $_->{server}->{address} . $_->{name}, channels), "\n";
+    my $out = { suppress => join("\0", map $_->{server}->{address} . $_->{name}, channels) };
```
ailin-nemui: how about you replace this with a structure / array instead of the awkward NUL byte

vague666:
That can be done yeah, didn't even think of it, tried to keep as much of the old script as possible but yeah, it makessense now that we use Storable

scripts/buf.pl
```diff
@@ -40,64 +43,59 @@
...
-
-    my @suppress = split /\0/, readline $fh;
-    if (settings_get_bool 'upgrade_suppress_join') {
-	chomp $suppress[-1];
-	@suppress{@suppress} = (2) x @suppress;
-    }
+    my $in = retrieve($fn) or die $!;
+    unlink $fn;
+  
+    my @suppress = split /\0/, $in->{suppress};
```
ailin-nemui: ... then you would need to remove this split here

scripts/buf.pl
```diff
@@ -40,64 +43,59 @@
...
+              push @$output, { level => $line->{info}{level}, data => $line->get_text(1) };
+              $line = $line->next;
+              redo if defined $line;
+            }
+        }
+        push @{$out->{windows}}, { refnum => $window->{refnum}, lines => $output };
     }
-    close $fh;
+    my $old_umask = umask 0077;
+    store($out, _filename) or die $!;
```
ailin-nemui: I can't find any mention in the docs that Storable sets $!. I think you would have to die with a generic error string ("storing failed")

scripts/buf.pl
```diff
@@ -40,64 +43,59 @@
...
+        next unless $window;
+        my @lines  = @{ $win->{lines} || [] };
+        next unless @lines;
+    
+        my $view = $window->view;
+        $view->remove_all_lines();
+        $view->redraw();
+        my $sep = settings_get_str 'upgrade_separator';
+        $sep .= "\n" if $sep ne '';
+        $window->gui_printtext_after(undef, MSGLEVEL_CLIENTNOTICE, "\cO$sep\n");
```
ailin-nemui: this separator notice should go LAST after all the other lines have been printed in each window, at least that's how it originally was

scripts/buf.pl
```diff
@@ -40,64 +43,59 @@
...
-    open my $fh, q{<}, $fn or die "open $fn: $!";
-    unlink $fn or warn "unlink $fn: $!";
-
-    my @suppress = split /\0/, readline $fh;
-    if (settings_get_bool 'upgrade_suppress_join') {
-	chomp $suppress[-1];
-	@suppress{@suppress} = (2) x @suppress;
-    }
+    my $in = retrieve($fn) or die $!;
+    unlink $fn;
```
ailin-nemui: why did you remove the warning here??

scripts/buf.pl
vague666: iirc the screen got scrolled up if I did it that way so the =Upgrade= wasn't visible

ailin-nemui: move it below the loop and use $window->last_line_insert instead of undef, should work I think

scripts/buf.pl
vague666: It must've gone out with speed, no solid reason as to why, I readded it :)

scripts/buf.pl
```diff
@@ -40,64 +42,60 @@
...
+        my $view = $window->view;
+        $view->remove_all_lines();
+        $view->redraw();
+        for my $line (@lines) {
+            my $level = $line->{level};
+            my $data  = $line->{data};
+            $window->gui_printtext_after($window->last_line_insert, $level, "$data\n");
+        }
+        my $sep = settings_get_str 'upgrade_separator';
+        $sep .= "\n" if $sep ne '';
```
ailin-nemui: this line seems to cause an empty line to appear on screen

ailin-nemui: maybe just remove it?

vague666: Maybe we should just remove it?

vague666: Or remove the \n from 
$window->gui_printtext_after($window->last_line_insert, MSGLEVEL_CLIENTNOTICE, "\cO$sep\n");

ailin-nemui: that seems buggy (bug was in the original script) if we have an empty $sep then we should not print anything (also not a \c0) otherwise we have an incomplete line hanging around in the buffer

vague666: Alright, that should be it :)

ailin-nemui:
you need to test "if $sep", defined is too strong. it will be defined but empty string

vague666:
That should do it, I used the defined to avoid $sep being '0', but didn't think of the consequences

ailin-nemui:
ok shall we merge it then?

vague666:
Maybe run a quick test first? It looked like you tested it as we went along here in the end

ailin-nemui:
please test it and tell me if everything is fine ^^

vague666:
And we are still using gui_printtext_after, should we change it to straight prnt instead? Do we have access to $Irssi::VERSION or whatever?

vague666:
Appears to work, it gets my thumbs up

PR #495, opened 2018-03-30
[mpd.pl] unescaped left brace in regex becomes an error in perl5 v26

yabbes:
I just escaped the braces in the regex statements as it becomes an error that lets the script crash in the current version of perl


ailin-nemui:
please increase the script version so that others can benefit from the fix

PR #497, opened 2018-04-03
[hilite_url] Make the url color configurable

schtibe:
Reaction to https://github.com/irssi/scripts.irssi.org/pull/329#commitcomment-20939744


ailin-nemui:
are you sure this works?

ailin-nemui:
ah right ok forget what I wrote, seems fine :-)

PR #500, opened 2018-04-10
[imdb] update connection to the website

bw1:


scripts/imdb.pl
```diff
@@ -41,11 +42,17 @@ sub event_nickchange {
 	else {
 
 	    # Fetch the movie detail page
-    	    my $req = new HTTP::Request(GET => "http://us.imdb.com/Title?$id");
```
ailin-nemui: this change is not strictly necessary as the redirect is there. but ok.

scripts/imdb.pl
```diff
@@ -41,11 +42,17 @@ sub event_nickchange {
...
 
 	    # Get the title and year from the fetched page
-    	    if($res->is_success && $res->content =~ /<title>(.+?) \((\d+)\)<\/title>/) {
+	# https://www.imdb.com/title/tt1234567/
+	# <title>&quot;So You Think You Can Dance&quot; The Top 14 Perform (TV Episode 2008) - IMDb</title>
+	# https://www.imdb.com/title/tt0234567/
+	# <title>The Ranchman's Nerve (1911) - IMDb</title>
+		#
+	    if($res->is_success
+		&& decode_entities($res->as_string) =~ /<title>(.+?) \((.+)\).*<\/title>/) {
```
ailin-nemui: it's not needed to change to as_string (does a different thing), and the wrong place to decode_entities. What has to be changed is solely the regexp format to catch the title content (it now has - imdb at the end)

scripts/imdb.pl
```diff
@@ -41,11 +42,17 @@ sub event_nickchange {
...
 
 		my ($title, $year) = ($1, $2);
```
ailin-nemui: the right place to decode entities is below here v--- . it's already in the code but in a bit too basic fashion

scripts/imdb.pl
ailin-nemui: you could evaluate decoded_content in case there is some specific charset encoding  applied on the web page

PR #502, opened 2018-04-27
[cddb] remove cddb.pl

bw1:


ailin-nemui:
just wondering, is freedb compatible to replace in this script?

bw1:
maybe 

but this script is a bot.

PR #503, opened 2018-04-28
[myimdb] remove myimdb.pl

bw1:
@timvw


timvw:
I will not miss it ;)

ailin-nemui:
why remove it? too lazy to fix :)?

PR #508, opened 2018-05-14
[findbot.pl] private message requests are allowed and changed filelist creation

hamsterbacke:
- requesting in a private message is allowed when user is in a monitored channel
- filelist creation script at the bottom of this script has been changed to support utf8, multiple folders, with or without file and 7z compression of the command list

### DIFF
--- a/scripts/findbot.pl
+++ b/scripts/findbot.pl
@@ -86,6 +86,8 @@
 ###############################################################################
 # CHANGES
 #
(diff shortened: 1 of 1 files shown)


ailin-nemui:
@henrik-andreasson any opinion here?

henrik-andreasson:
@ailin-nemui I'm not associated with this project since 10 years +++

PR #509, opened 2018-05-20
[timer.pl] Add timer setting to disable stop messages

PierrotMerlot:


ailin-nemui:
@kke

kke:
Heh, it's still alive :D I don't think there's any of my code left in it.

PR #516, opened 2018-07-03
[oops] Also call /names on ll

mineo:
ll is aliased to `ls -l` on a lot of distributions by default and has made its
way into (at least my) muscle memory right next to regular ls.

### DIFF
--- a/scripts/oops.pl
+++ b/scripts/oops.pl
@@ -2,32 +2,32 @@ use strict;
 use vars qw($VERSION %IRSSI);
 
 use Irssi;
-$VERSION = '20071209';
+$VERSION = '20180707';
 %IRSSI = (
     authors     => '',
     contact     => '',
     name        => 'oops',
     description =>
-'turns \'ll\' and \'ls\' in the beginning of a sent line into the names or whois commands',
+    'turns \'ll\' and \'ls\' in the beginning of a sent line into the names or whois commands',
     license => 'Public Domain',
-);
+    );
 
 sub send_text {
-    my $pattern = qr/(^ll |^ll$|^ls |^ls$)/;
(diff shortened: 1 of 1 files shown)


vague666:
I don't really understand why sent _text_ is checked, I would instead use an alias /alias ll /names and /alias ls /whois $T, or signal_add('send command' => ...); but if must use 'send text' then I suggest these changes:
I would use 
<pre>
my $pattern = qr/^(ll|ls)\s*$/;
if($data =~ $pattern) { if($witem->{type} eq "QUERY") { ... } elsif(...) { ... } }
</pre>
or
<pre>
my $pattern = qr/^(ll|ls)\s*$/;
%cmds = (
  QUERY => sub { Irssi::command(sprintf("wii %s", Irssi::parse_special('$T'))) },
  CHANNEL => sub { Irssi::active_win->{active}->command('names') }
);
if($data =~ $pattern) { $cmds{$witem->{type}} && Irssi::signal_stop }
</pre>

dequis:
>I don't really understand why sent text is checked

I thought the same thing at first, but this is when you don't type the slash, just send the word "ls" to the channel. I don't know how I feel about modifying the script to add aliases or the fact that this is all the script did.

mineo:
I've implemented the suggestions by @vague666.

I can't say anything about whether the script makes sense for people not using it - I like it a lot, but it probably looks a bit like [XKCD 1172](https://xkcd.com/1172/) to you.

ailin-nemui:
I wonder if it would confuse other users who want to have names on ls but not on ll when they update their scripts

ailin-nemui:
I notice that the script has now a different behaviour, previously it would ignore anything starting with `ls `, now it only ignores exactly `ls`

bw1:
see #594

PR #518, opened 2018-07-13
[active_notice] version 15.01.01

bw1:
@GeertHauwaerts

### DIFF
--- a/scripts/active_notice.pl
+++ b/scripts/active_notice.pl
@@ -2,61 +2,201 @@
 
 ## Bugreports and Licence disclaimer.
 #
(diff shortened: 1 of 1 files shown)


scripts/active_notice.pl
```diff
@@ -2,61 +2,201 @@
...
+        # Print the notice in the active window.
+        ##
+        
+        $awin->print($text, MSGLEVEL_NOTICES);
+    }
+}
 
-    return if $witem;
 
-    $text =~ s/%/%%/g;
```
ailin-nemui: it seems to me like this fix %/%% may be missing from the "new" version

PR #523, opened 2018-08-21
[usercount.pl] fix crash when server doesn't have isupport

ailin-nemui:


dequis:
Hm, shouldn't we provide generic/empty defaults for servers like this?

PR #524, opened 2018-08-28
Feature update for slack_complete.pl

ossih:
* Support for common mentions (@here, @channel, @everyone)
* Use completion_char setting instead of hard coded colon
* Option to not add completion_char when completing first word of current line


ailin-nemui:
@mortenlj

PR #525, opened 2018-09-01
[dancer_hide_477] remove dancer_hide_477.pl

bw1:
### dancer-ircd
* =< 2005 [Freenode](https://en.wikipedia.org/wiki/Freenode#Server_software)
* =< 2013 [debian](https://tracker.debian.org/pkg/dancer-ircd)
* [wiki](https://en.wikipedia.org/wiki/Comparison_of_Internet_Relay_Chat_daemons)

@GeertHauwaerts


dequis:
477 is this in ircd-seven:

```
":%s 477 %s %s :Cannot join channel (+r) - you need to be identified with services"
```

Not sure there are reasons to hide this. The script limits to servers identified as dancer. It could always be replaced with trigger.pl with `-stop`, i think.

+1 for deleting i guess.

PR #528, opened 2018-09-07
added urlwindow.pl script

zdleaf:


ailin-nemui:
thanks for your contribution, can you please make your script work with
```
use strict;
```

zdleaf:
thanks - use strict; added

vague666:
Maybe use Regexp::Common and $RE{URI} to detect urls

PR #536, opened 2018-09-17
Rewrite code to work with window_default_hidelevel and other improvements

vague666:


bw1:
### old warnings
```
-!- Irssi: warning settings_get(window_default_hidelevel) : not found
-!- Irssi: Loaded script toggle_hidelevel
```
/show_levels_all , /hide_levels_all and/window togglelevel 
```
Use of uninitialized value in subroutine entry at 
/home/robert/versuche/irssi/current/./config/scripts/toggle_hidelevel.pl line 44.
```

vague666:
The warning is expected on pre 1.2.0, I don't know how else to check for the existence of a setting
But I'm not seeing the errors on line 44, try adding `|| 'NONE'` to line 43?

vague666:
I'm pretty sure I tested the version string method and at some point it must've worked, but changing the version now doesn't seem to do anything. I've seen some script use a date string instead so maybe I'll go with that instead

PR #537, opened 2018-09-20
[duckduckgo] add duckduckgo.pl

bw1:


vague666:
You really shouldn't parse html with regular expressions. Mojolicious makes this easy and has a built-in eventloop

vague666:
I wrote a simple sample script, make what you want out of it :)
https://vague.se/pastes/ddg_async.pl

bw1:
nice
but the ddg function blocks irssi

vague666:
Which is why I wrote do what you want out of it :) it needs to fork or be run through something like IO::Async::Loop::Glib

PR #554, opened 2018-11-24
[server_time] sneak in server-time extension.

itsjohncs:
This script implements IRCv3's server-time extension for most users'
purposes. For example, ZNC users can use this plugin to get clean
playback buffers.

I've updated this script to drop a dependency I had to download from
CPAN when I started using it. I did not initially create this script,
but I can provide support going forward. My repo for issues is at
https://github.com/itsjohncs/irssi-server-time (mentioned in the
script as well).

I plan to implement the server-time extension in irssi's core myself
eventually, so maintaining this script seems fitting.

### DIFF
--- a/scripts/server_time.pl
+++ b/scripts/server_time.pl
@@ -63,7 +63,7 @@ our %IRSSI = (
     name        => 'server_time',
     description => 'Implements the IRCv3 "server-time" capability',
     license     => 'MIT',
-    url         => 'https://github.com/brownhead/irssi-server-time',
+    url         => 'https://github.com/itsjohncs/irssi-server-time',
 );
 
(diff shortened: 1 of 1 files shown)


dwfreed:
This will have interesting side effects, like depending on when the statusbar gets redrawn, the clock may appear to go back in time (because it also uses timestamp_format, and you're changing the format to a fixed string).

An improvement idea: instead of prepending the date on the line if it's different than the current date, just print the "Day changed" format and leave the timestamp format alone (I have the date in my timestamp format already, so this would result in duplication anyway).

brownhead:
@dwfreed what is the "day changed" format?

PR #556, opened 2018-11-25
[away.pl] Update code to work with autoaway

vague666:


ailin-nemui:
any particular reason to change the settings names?

ailin-nemui:
does this even work o_O now we have away_reason registered for 2 categories, I bet there is a bug in irssi that makes this a bad idea

PR #561, opened 2018-12-16
[trackbar] fix bug

bw1:
*  #559
  *  remove redraws


ailin-nemui:
thanks for the bug fix! however, I want to try something else instead:

is it possible, to detect

if the line was actually written to the textbuffer?

if this can be detected, then we should not set the bookmark.

bw1:
#565

PR #571, opened 2019-01-07
[schwaebisch] utf8

bw1:


ailin-nemui:
it may be better to query the charset setting in each command, lest the user changes it

bw1:
ok

ailin-nemui:
mhm, it's actually a funny case having to bind different commands.... 

I doubt it's good to run command_bind more than once

maybe we should stick to your original version

bw1:
ok

PR #573, opened 2019-01-13
[whos] remove whos.pl

bw1:
negatives
  *  the same as "/who #channel" 
  *  "/whos"  filter the result by server tag result is nothing
  *  "/whos .*"  give the result of /who #channel


ailin-nemui:
it could be used to filter who is on some server

example:

`/whos adams`

-> show only people who are on the "adams" server of freenode

PR #575, opened 2019-01-14
[ircsec] view the missing of blowfish

bw1:


ailin-nemui:
is this hard dependency or will it autoload perl blowfish module as long as it is installed?

bw1:
autoload by Crypt::CBC via default config
and eval hiden the error message

```
60      eval {
61         $cipher = Crypt::CBC->new( -key             => $key,
62                                    -cipher          => $algo,
63                                    -iv              => '$KJh#(}q',
64                                    -literal_key     => 0,
65                                    -padding         => 'space',
66                                    -header          => 'randomiv'
67                                  );
68  
69      };
```

```
199 Irssi::settings_add_str($IRSSI{name}, 'ircsec_default_cipher', 'Blowfish');
```

ailin-nemui:
`use Crypt::Blowfish` is NOT required

instead the error needs to be un-hidden so people know they must install it (if it is missing)

bw1:
eval is removed

PR #577, opened 2019-01-21
ignore_join_blob: ignore text displayed when (re)joining a channel

trapd00r:
This script solves the issue [described here](https://github.com/irssi/irssi/issues/992) where the text displayed on join is unignorable despite them being of the level CRAP.

### DIFF
--- a/scripts/ignore_join_blob.pl
+++ b/scripts/ignore_join_blob.pl
@@ -1,4 +1,4 @@
-# vim:ft=perl:et:
+# vim:ft=perl:et:sw=2:ts=2:
 use strict;
 use Irssi;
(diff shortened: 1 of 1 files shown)


ailin-nemui:
this might be very specific to your theme...

trapd00r:
Is it? Tried with default and it worked fine.

scripts/ignore_join_blob.pl
```diff
@@ -0,0 +1,35 @@
+# vim:ft=perl:et:
```
bw1: # vim:ft=perl:et:sw=2:ts=2:

PR #581, opened 2019-02-06
[on][crit] new option

bw1:
@dgl

### DIFF
--- a/scripts/on.pl
+++ b/scripts/on.pl
@@ -1,5 +1,10 @@
+use strict;
 use Irssi 20011210.0000 ();
-$VERSION = "1.12";
+use Storable;
+
+use vars qw/$VERSION %IRSSI/;
+
+$VERSION = "1.13";
 %IRSSI = (
     authors     => 'David Leadbeater',
     contact     => 'dgl@dgl.cx',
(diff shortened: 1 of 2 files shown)


ailin-nemui:
it would be better if you can keep the old format, for backward compatibility

bw1:
### old:
```
/on add 332 /echo hallo welt
/on add 332 -perl print "hallo perl";
/on add 332 -server /echo hallo welt
/on add 332 -channel #test /echo hallo welt
/on add 332 -stop /echo hallo welt
/on add 332 -perl -stop print "hallo perl";
```

### on.save
```
event 332^@all^@^@^@/echo hallo welt
event 332^@all^@1^@^@print "hallo perl";
event 332^@mylocalnet^@^@^@/echo hallo welt
event 332^@all^@^@#test^@/echo hallo welt
event 332^@all^@1^@^@/echo hallo welt
event 332^@mylocalnet^@1^@^@/echo hallo welt
event 332^@all^@11^@^@print "hallo perl";
```

-stop and -perl is the same in the file

scripts/on.pl
```diff
@@ -138,8 +133,11 @@ sub signal_handler {
    my($item, @stuff) = @_;
    my $signal = Irssi::signal_get_emitted();
 
+
    if(exists $on{$signal}) {
       for(@{$on{$signal}}) {
+         debug(Dumper($_));
```
ailin-nemui: I think there is error here...

bw1:
:-/

done

PR #583, opened 2019-02-09
[whos] connected server as regex

bw1:


ailin-nemui:
Irssi::print is not nice, is it for debug only? -> remove; or should it be helpful info -> make it formatable

by the way, this script can conflict if you use /whos on 2 servers at the same time (would need to make SERVER_NAME & results depend on server tag)

bw1:
better?

scripts/whos.pl
```diff
@@ -70,14 +77,32 @@ sub sig_whos
...
+   $res{$server->{tag}}->{result}->{$who[5]}=[@who];
 }
 
 sub sig_whosend
 {
    my ($server, $msg, $nick, $address, $target) = @_;
+   if ($res{$server->{tag}}->{regex} eq '') {
+      $res{$server->{tag}}->{regex}= $nick;
+   }
+   Irssi::print("%9regex:%n".$res{$server->{tag}}->{regex},MSGLEVEL_CRAP);
```
ailin-nemui: if you want to remain this line, use "Irssi::printformat" and register a format for it

bw1:
done

PR #592, opened 2019-02-19
[file] Getopt::Long

bw1:
@dgl


scripts/file.pl
```diff
@@ -19,7 +20,7 @@
 -echo abuses a bug in the script and is useful for testing
```
dgl: Either remove this note (but I find -echo useful and had forgotten it relies on side effects) or consider generically treating unknown options as commands...

bw1:
done

PR #603, opened 2019-02-26
go.pl: fix short circuiting during channel search

majuscule:
example: currently `/go #vim` when `#vim` and `#vim-ale` are both open will incorrectly go to `#vim-ale` depending on the order of your windows. this fixes that.

### DIFF
--- a/scripts/go.pl
+++ b/scripts/go.pl
@@ -29,7 +29,7 @@ use Irssi::Irc;
 #     completion. The leading '#' of channel names is optional either way.
 #
 
-$VERSION = '1.1';
+$VERSION = '1.1.1';
 
 %IRSSI = (
     authors     => 'nohar',
@@ -37,7 +37,7 @@ $VERSION = '1.1';
     name        => 'go to window',
     description => 'Implements /go command that activates a window given a name/partial name. It features a nice completion.',
     license     => 'GPLv2 or later',
-    changed     => '2017-02-02'
+    changed     => '2019-02-25'
 );
 
 sub _make_regexp {
@@ -81,12 +81,18 @@ sub cmd_go
 		Irssi::settings_get_bool('go_match_case_sensitive'),
 		Irssi::settings_get_bool('go_match_anchored'));
 
+	my @matches = [];
 	foreach my $w (Irssi::windows) {
 		my $name = $w->get_active_name();
-		if ($name =~ $re) {
+		if ($name =~ /$re$/) {
 			$w->set_active();
 			return;
+		} elsif ($name =~ /$re/) {
+			push(@matches, $w);
 		}
+	}
+	if (@matches) {
+		$matches[0]->set_active()
 	}
 }
 
@@ -103,3 +109,5 @@ Irssi::settings_add_bool('go', 'go_complete_anchored', 0);
 #   - made case-sensitivity of match configurable
 #   - made anchoring of search strings configurable
 #
+# 2019-02-025  1.1.1  dylan lloyd <dylan@disinclined.org>
+#   - fixed short-circuiting during channel search
--- a/scripts/go.pl
+++ b/scripts/go.pl
@@ -110,4 +110,4 @@ Irssi::settings_add_bool('go', 'go_complete_anchored', 0);
 #   - made anchoring of search strings configurable
 #
 # 2019-02-025  1.1.1  dylan lloyd <dylan@disinclined.org>
-#   - fixed short-circuiting during channel search
+#   - prefer exact channel matches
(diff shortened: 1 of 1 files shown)


majuscule:
Just read from the README, `Do not create explicit minor revision numbers if they are not present (1.3 becomes 1.4 or 2.1, not 1.3.1)`, happy to change this if desired but I do not understand the reasoning. Additionally, 1.0.x is a patch release, not a minor revision. https://semver.org/

ailin-nemui:
this will prefer channels that match at the end, eg. /go m goes to vim instead of mutt

bw1:
```
/run go.pl
/go tes
09:42 -!- Irssi: Error in script go:
09:42 Can't call method "set_active" on unblessed reference at 
          /home/robert/versuche/irssi/current/./config/scripts/go.pl line 95.
```

```
32  $VERSION = '1.1.1';
                  ^^^
```

```
112 # 2019-02-025  1.1.1  dylan lloyd <dylan@disinclined.org>
              ^      ^^^
```

scripts/go.pl
```diff
@@ -81,13 +81,19 @@ sub cmd_go
 		Irssi::settings_get_bool('go_match_case_sensitive'),
 		Irssi::settings_get_bool('go_match_anchored'));
 
+	my @matches = [];
 	foreach my $w (Irssi::windows) {
 		my $name = $w->get_active_name();
-		if ($name =~ $re) {
+		if ($name =~ /$re$/) {
```
vague666: Better make this if(fc $name eq fc $chan) {

majuscule: That would be different behavior from the partial match then. I suppose I could add it to both. I don't think I've ever seen an IRC channel w/ caps anyway though.

majuscule:
@bw1 I have fixed the error, my perl is a bit rusty :-). Would you please address my comments re versioning?

@ailin-nemui I believe you are mistaken. I have tested this behavior and it goes to #mutt as expected.

bw1:
``` perl
90  		} elsif ($name =~ /$re/) {
91  			push(@matches, $w);
92  		}
93  	}
94  	if (@matches) {
95  		$matches[0]->set_active();
96  	}

```



``` perl
90  		} elsif ($name =~ /$re/) {
91  			$w->set_active();
92  			return;
93  		}
```

equal or not equal

scripts/go.pl
vague666: How is line 87 supposed to work according to you? To me it looks like it should go to an exactly matched channel, as Nei wrote, currently it matches from the end, ie /go tt would go to #mutt instead of #ttcars

bw1:
```
$ $s='#test';
#test
$ $re=_make_regexp('est',1,1);
^#?est
$ $s =~ /$re$/;
$ $s =~ /$re/;
```

``` bash
$ $s='#test';
#test
$ $re=_make_regexp('est',1,0);
est
$ $s =~ /$re$/;
1
$ $s =~ /$re/;
1
```

```
$ $s='#test';
#test
$ $re=_make_regexp('es',1,0);
es
$ $s =~ /$re$/;
$ $s =~ /$re/;
1
```

powert by re.pl

ailin-nemui:
let's try to rethink this PR and decide on the following question:

- think prefer perfect match could make sense. does it need a setting or can it be "always on"?

As for the implementation I guess it could be done with a regex that is anchored to start and end

- if it goes to vim or mutt depends on the /set go_match_anchored off. in general I think anchoring to the end is **bad** so we should fix this

- for the special case that go_match_anchored is ON (default), additional anchor to the end will be "almost" like perfect match (plus the optional #)

bw1:
> example: currently `/go #vim` when `#vim` and `#vim-ale` are both open will incorrectly go to `#vim-ale` depending on the order of your windows. this fixes that.

### test this
```
84  	my $active='';
85  	if ( defined $witem ) {
86  		$active= $witem->window()->get_active_name();
87  	}
88  	foreach my $w (Irssi::windows) {
89  		my $name = $w->get_active_name();
90  		if ($name =~ /$re/) {
91  			if ( $name ne $active  ) {
92  				$w->set_active();
93  				return;
94  			}
95  		} 
96  	}
```

scripts/go.pl
majuscule: Sorry, I misunderstood the point of your original suggestion. I thought that you were just aiming for case insensitivity. I have pushed a new commit with this code amended as per your proposal.

majuscule:
@bw1 I have updated the changeset per @vague666's suggestion, hopefully this resolves your concerns. I am not sure what you meant re "test this".

@ailin-nemui I feel strongly that exact match should be default behavior.

bw1:
@majuscule 
[try this](https://github.com/bw1/gotest/blob/master/gotest.pl)

majuscule:
Fixed case sensitivity option handling. The anchoring is otherwise still handled by `_make_regexp`. If there are any other issues would you please point them out explicitly? Thanks!

ailin-nemui:
thanks for your contribution @majuscule

PR #604, opened 2019-02-27
[scripthelp] bug fix

bw1:


ailin-nemui:
mhm, this would be a nice idea to have built in to irssi :thinking:

PR #608, opened 2019-03-03
[dice_concise] replace 'when' and 'given' with 'if'

bw1:


ailin-nemui:
smart match is deprecated, you should rewrite the script to use non smart match

bw1:
done

ailin-nemui:
@Makaze

Makaze:
The changes look solid.

PR #609, opened 2019-03-06
add 'tags' column in the website

bw1:


scripts/wordcompletition.pl
```diff
@@ -8,12 +8,14 @@
...
-    description => "Adds words from IRC to your tab-completion list",
+    name        => "wordcompletition",
+    description => "Adds words from IRC to your tab-completion list".
+					" and store in a mysql database.",
     license     => "Public Domain",
     url         => "http://midgard.liu.se/~n02jesli/perl/",
     changed     => "2017-03-19",
-	modules     => "DBD::SQLite"
+    modules     => "DBD::SQLite",
+    tags        => ['completition','database'],
```
vague666: typo

bw1: thx

ailin-nemui:
interesting idea, we need to evaluate this in depth. I'm not convinced about (1)

- do we need/want this?
- who should be in charge of providing the tags?
- should the tags be in the script source or external?

as long as you have the right "tags" in your description, the search/filter will also be able to find the script

@irssi/developers @irssi/script-reviewers

PR #610, opened 2019-03-10
[rud_emotes] Added utf-8 support and auto-completion.

rud01f:
- fixed messages containing utf-8 characters
- added tab-completion to emotes
- added "rud_emotes_space_after_completion" settings flag


bw1:
```
74  	changed		=> 'Mon Nov 07 14:54:38 2016'
```
:octocat:

Issue #569, opened 2019-01-01
[lastspoke.pl] permanent storage, maybe text file


laama1:
This script should have a permantent storage that would stay after reboot. This way the script would be much more useful.

ailin-nemui:
@laama1 does the work done by bw1 in #613 work for you?

PR #613, opened 2019-03-17
[lastspoke] #569 and utf8

bw1:


ailin-nemui:
I'm worried because 

- not every channel uses utf8

bw1:
done

scripts/lastspoke.pl
```diff
@@ -20,11 +20,14 @@
 # Triggers on !lastspoke <nick>, !seen <nick> and !lastseen <nick>
 # 
 use strict;
+use utf8;
+use Encode qw/decode encode/;
 use Irssi;
 use Irssi::Irc;
+use JSON::PP;
```
laama1: Requires installing of extra module perl-JSON-PP, at least on my fedora 31 server system. I installed it and am now testing this script.

ailin-nemui:
@sndrsmnk

sndrsmnk:
Ghe. I'm not using this script myself, anymore. Cool to see it live on. ;)

I would personally not use JSON::PP, it requires 'external' modules on almost all systems and we're not dealing with large data structures where. Some Data::Dumper & eval (evil?:P) trickery would suffice:

Eg save():
```
if (open(FD, ">dump.perlObj")) {
  print FD Dumper(%lasthash);
  close(FD);
}
```

Eg load():
```
if (open(FD, "<dump.perlObj")) {
  local $/ = undef; my $blob = <FD>; close(FD);
  %lasthash = eval 'my ' . $blob;
}
```

As for the utf-8 decode/encode stuff, why bother?

Just assume everything is utf-8 instead of encoding the eventual output to whatever your terminal is set to.

bw1:
a new version

PR #614, opened 2019-03-19
[leodict] new connector

bw1:


scripts/leodict.pl
```diff
@@ -4,27 +4,44 @@
 use strict;
 
 use vars qw($VERSION %IRSSI);
-$VERSION = '20040515';
+$VERSION = '2.0';
```
vague666: might not work to change version like that, scriptassist will get confused or not know there is a new version

scripts/leodict.pl
```diff
@@ -53,41 +89,62 @@ ()
...
+	    my $r=$1;
+	    my @col=();
+	    my $c=0;
+	    while ( $r =~ m#(<td.*?</td>)#p ) {
+		$r=$';
+		$_=$1;
+		$_=~s/<.*?>//g;
+		if (length($_) >2) {
+		    $col[$c]=$_;
+		    $c++;
```
vague666: Don't parse html with regex, there are html parsers for that
perlbot ┃ vague: Don't parse or modify html with regular expressions! See Mojo::DOM or HTML::TreeBuilder or suchlike. For a Feb-2014 coverage of main techniques see: http://is.gd/4FwOKu . If your response includes "that's overkill. i only want to..." you are wrong.           http://enwp.org/Chomsky_hierarchy and http://is.gd/Yk3uww for why not to use regex on HTML. | https://pbs.twimg.com/media/CgjfnfeW0AA0G5h.jpg

scripts/leodict.pl
bw1:
```
$ compare_versions('1','2');         
older                                                                           
$ compare_versions('20040515','2.0');
newer
```
:(

##### README.markdown
##### To increase the version numbering, take the following conditions into account:

  1.    Do not create explicit minor revision numbers if they are not present (1.3 becomes 1.4 or 2.1, not 1.3.1)
  2.    When modifying an existing feature, increase the minor revision number (1.3 becomes 1.4, not 2.1)
  3.    When adding a new feature, increase the major revision number (1.3 becomes 2.1, not 1.4)

vague666: no idea what compare_versions returns :) but
`perl -Mversion -E 'say version->parse("20040515") > version->parse("2.0") ? "true" : "false"'`
`true`

bw1: ####  scriptassist.pl
```
912 sub compare_versions {
913     my ($ver1, $ver2) = @_;
914     for ($ver1, $ver2) {
915 	$_ = "0:$_" unless /:/;
916     }
917     my @ver1 = split /[.:]/, $ver1;
918     my @ver2 = split /[.:]/, $ver2;
919     my $cmp = 0;
920     ### Special thanks to Clemens Heidinger
921     no warnings 'uninitialized';
922     $cmp ||= $ver1[$_] <=> $ver2[$_] || $ver1[$_] cmp $ver2[$_] for 0..scalar(@ver2);
923     return 'newer' if $cmp == 1;
924     return 'older' if $cmp == -1;
925     return 'equal';
926 }
```
:construction_worker_man: 
:hammer_and_pick:

PR #616, opened 2019-03-21
[leodict] new connector

bw1:
with Mojo

### DIFF
--- a/scripts/leodict.pl
+++ b/scripts/leodict.pl
@@ -4,27 +4,45 @@
 use strict;
 
 use vars qw($VERSION %IRSSI);
-$VERSION = '20040515';
+$VERSION = '20180321';
 %IRSSI = (
-    authors     => 'Stefan \'tommie\' Tomanek',
-    contact     => 'stefan@pico.ruhr.de',
+    authors     => 'Stefan \'tommie\' Tomanek, bw1',
+    contact     => 'bw1@aol.at',
     name        => 'leodict',
     description => 'translates via dict.leo.org',
     license     => 'GPLv2',
     url         => 'http://irssi.org/scripts/',
-    changed     => $VERSION,
-    modules     => 'LWP::Simple Data::Dumper',
+    modules     => 'Mojo::UserAgent Encode JSON::PP Mojo::DOM Getopt::Long POSIX',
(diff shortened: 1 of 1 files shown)


scripts/leodict.pl
```diff
@@ -4,27 +4,45 @@
...
-    authors     => 'Stefan \'tommie\' Tomanek',
-    contact     => 'stefan@pico.ruhr.de',
+    authors     => 'Stefan \'tommie\' Tomanek, bw1',
+    contact     => 'bw1@aol.at',
     name        => 'leodict',
     description => 'translates via dict.leo.org',
     license     => 'GPLv2',
     url         => 'http://irssi.org/scripts/',
-    changed     => $VERSION,
     modules     => 'LWP::Simple Data::Dumper',
```
vague666: Mojo::DOM missing, but if you are going to use Mojo then using the whole mojo suite is just some steps away, use Mojo::UserAgent to fetch the data, say $ua->get('www.perl.org')->result->dom->at('title')->text; result->json gives json output also, not so many dependencies with just cpanm Mojolicious :)
* it's only suggestions so do what you prefer and feel comfortable with :) *

ailin-nemui:
I think you can change from LWP::Simple to Mojo::UserAgent `get'

bw1:
done

scripts/leodict.pl
```diff
@@ -4,27 +4,45 @@
...
-    contact     => 'stefan@pico.ruhr.de',
+    authors     => 'Stefan \'tommie\' Tomanek, bw1',
+    contact     => 'bw1@aol.at',
     name        => 'leodict',
     description => 'translates via dict.leo.org',
     license     => 'GPLv2',
     url         => 'http://irssi.org/scripts/',
-    changed     => $VERSION,
-    modules     => 'LWP::Simple Data::Dumper',
+    modules     => 'LWP::Simple Encode JSON::PP Mojo::DOM Getopt::Long POSIX',
```
ailin-nemui: now we can remove lwp

bw1:
done

vague666:
Also, JSON::PP shouldn't be needed since Mojo does json with Mojo::JSON :)

PR #617, opened 2019-04-14
[bitlbee_timestamp] print time with level

bw1:
@dequis
@timing


ailin-nemui:
I just had a quick glance but this may be too radical. Levels influence logging and setting activity bypasses some ignorance rules.

It certainly goes in the right direction but we should be careful before merging this. It may be we have to change something in Irssi core first

PR #632, opened 2019-05-09
[elist] Add optional -yes/-YES option

### LINKED ISSUE #630: [elist.pl] List command refuses to run without -YES option
The `/LIST` command requires the `-YES` option to confirm the user wants a large amount of data to be sent to the client. Currently, `elist.pl` does not include this option, so attempting to use it prints a warning message and does not list any channels.

Running `/elist min=1000` on Freenode gives:
```
[Freenode] <--> elist (min=500, max=10000, mask='')
-!- Irssi: Doing this is not a good idea. Add -YES option to command if you really mean it
```

I was unable to find any way to disable this warning in the Irssi docs, please let me know if I am missing something. I've confirmed changing `$server->command("LIST " ...);` to `$server->command("LIST -YES " ...);` on line 81 resolves this issue. If this is an acceptable solution, I would be happy to submit a pull request for this.

jonpavelich:
Adds an optional argument (`-yes`/`-YES`) that allows users to pass `-YES` to the underlying `/LIST` command when required. Irssi warns the user and prompts for this option when required.

Fixes #630

### DIFF
--- a/scripts/elist.pl
+++ b/scripts/elist.pl
@@ -42,6 +42,7 @@ sub elist_config_init {
   %elist_config = (
     mincount => 0,
     maxcount => 10000,
+    yes => "",
     chanmask => ""
   );
 }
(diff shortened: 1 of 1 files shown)


ailin-nemui:
please increase the version number :-)

jonpavelich:
> please increase the version number :-)

Done :-)

PR #640, opened 2019-06-07
[example.pl] add example.pl

bw1:


scripts/example.pl
```diff
@@ -0,0 +1,71 @@
...
+	$test_str= Irssi::settings_get_str($IRSSI{name}.'_test_str');
+}
+
+Irssi::theme_register([
+	'example_theme', '{hilight $0} $1 {error $2}',
+]);
+
+Irssi::signal_add('setup changed', \&sig_setup_changed);
+
+Irssi::settings_add_str($IRSSI{name} ,$IRSSI{name}.'_test_str', 'hello world!');
```
madduck, line 66: shouldn't the 'hello world!' be `$test_str` instead, and the default set by initialising this variable?

bw1: `settings_add_str` let add irssi a setting with default value 'hello world!'

`sub sig_setup_changed' read out settings and maybe check values or react on chance of value

`$test_str` is for working in script (value is checked and change is already done)
init of this scalar is on last line

madduck: Hah, makes sense. I never stop learning, and I think I got it now. So please disregard my comment.

PR #650, opened 2019-07-11
[fnotify] Add own messages to fnotify writes

mrda:
fnotify allows you to do things like turn on blinkenlights when you are
hilighted in messages. Adding your own messages allows you to turn off
any local notifications once you respond.

Sample of how to do this is at
https://github.com/mrda/junkcode/blob/master/irssi-notify.sh

### DIFF
--- a/scripts/fnotify.pl
+++ b/scripts/fnotify.pl
@@ -7,7 +7,7 @@ $VERSION = '0.0.7';
 %IRSSI = (
 	name => 'fnotify',
 	authors => 'Tyler Abair, Thorsten Leemhuis, James Shubin' .
-               ', Serge van Ginderachter',
+               ', Serge van Ginderachter, Michael Davies',
 	contact => 'fedora@leemhuis.info, serge@vanginderachter.be',
 	description => 'Write notifications to a file in a consistent format.',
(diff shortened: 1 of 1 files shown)


bw1:
@purpleidea

bw1:
it would be an asset
a config option for the new behavior

ailin-nemui:
@mrda do you think you could make this an option (like /set fnotify_showself ON) ?

mrda:
@ailin-nemui Can you point me at an example of how to create a config value?  Thanks.

bw1:
[guide](https://github.com/shabble/irssi-docs/wiki/Guide#Use-Settings-for-Customisation)

ailin-nemui:
should be like this:

before signal_add:
```
Irssi::settings_add_bool('misc', 'fnotify_own', 0);
```

and in your signal handler, use:
```
if (Irssi::settings_get_bool('fnotify_own')) {
    ...
}
```

and to enable your custom feature:
```
/set fnotify_own ON
```
in Irssi

bw1:
Irssi::settings_get_bool(key)
Irssi::settings_add_bool(section, key, def)

bw1:
thx

PR #651, opened 2019-07-12
[mangle] extended

bw1:


scripts/mangle.pl
```diff
@@ -10,6 +10,7 @@
 use Irssi::TextUI;
 use POSIX;
 use Data::Dumper;
+use Irssi::TextUI;
```
guntbert: seems superfluous - see first line

PR #654, opened 2019-07-20
[theme] v0.03, theme.yaml, xterm fg- & bg-color

bw1:


ailin-nemui:
I think it would be better, if the script does not set the colour automatically!

make:

`/theme -setcolor|-s`

to set the colour, and `print "to set the colour as in screenshot, type /theme -s or configure your terminal with background colour: #abcdef ; text colour : #012345 "`

and maybe an error:
```
/theme -s
--> "Do not know how to set colour for your terminal, manually configure it for background : #...... ; text colour : #......"
```

and if you want `/set theme_autocolor ON|OFF` (default off)

PR #660, opened 2019-08-05
[quakequit] add a note

bw1:


ailin-nemui:
the note is not relevant for script users :-)

PR #665, opened 2019-08-23
[notes.pl] Use $_ inside a block instead

vague666:


ailin-nemui:
irssi/irssi#1112

ailin-nemui:
this has been fixed **in irssi**

PR #677, opened 2019-11-13
[dtach_away] add support for dtach

anarcat:
This forks and rewrites the `screen_away.pl` script into a new `dtach_away.pl` one. It's kind of hackish: maybe it would be best to add support for dtach into the existing script, but I found it easier to proceed that way because the capabilities of the two programs are not the same (save the executable bit on the socket, see https://github.com/crigler/dtach/issues/17#issuecomment-553577118).

This drops the socket autodetection. There's no equivalent to `screen -ls` in dtach, so that doesn't work. We could do process tracing with lsof to figure it out, but that seems horrible and it seems like minimal config to ask the user for the socket path.

Works for me anyways, I always use the same socket for IRC.

### DIFF
--- a/scripts/dtach_away.pl
+++ b/scripts/dtach_away.pl
@@ -6,7 +6,8 @@ use vars qw($VERSION %IRSSI);
 
 $VERSION = "1.0";
 %IRSSI = (
-    authors     => 'Andreas \'ads\' Scherbaum <ads@wars-nicht.de>',
+    authors     => 'Antoine Beaupré',
+    contact     => 'anarcat@debian.org',
     name        => 'dtach_away',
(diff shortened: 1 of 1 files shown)


bw1:
please change
``` perl
    authors	=> 'noname',
    contact	=> 'noname@example.org',
```
and a note of right dtach version 
(Version 0.8 has no executable bit)

anarcat:
> please change ...

do you mean like 93be31ee?

> and a note of right dtach version

i figured it wasn't quite worth saying we need a version of dtach that's less than 5 years old but okay, 977bc25f should do it. :)

anarcat:
yay thanks!

PR #712, opened 2020-05-16
mh_secureonlychan_489 v1.00

mh-source:
For users who are forced/prefer to use an older version of Irssi but still want this little issue fixed.

Fix for Irssi not fully supporting numeric 489 ERR_SECUREONLYCHAN.

see https://github.com/irssi/irssi/issues/1193 for details about the issue.

This script will display an error message in the status-window if you try to join a secure only channel without a secure connection. If you have the setting autoclose_windows off, you will still get the old raw message in the channel-window too.

The script will die gracefully with a message if loaded in a version of Irssi that already have the source-code fixed (ABI version 29 or higher).

see https://github.com/irssi/irssi/pull/1196 for details about the code fix.

### DIFF
--- a/_testing/config.yml
+++ b/_testing/config.yml
@@ -37,6 +37,7 @@ whitelist:
   - ixmmsa
   - l33tmusic
   - log2ansi
+  - mh_secureonlychan_489
   - trackbar22
   - tvmusor
   - xmmsinfo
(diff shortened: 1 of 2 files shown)


ailin-nemui:
thanks! please add your script to the white list in https://github.com/irssi/scripts.irssi.org/blob/master/_testing/config.yml

PR #716, opened 2020-07-13
Update quitrand.pl

simcop2387:
Clean up the perl a little.  Remove an uneccesary loop, use a lexical file handle to avoid possible conflicts, and use chomp instead of chop.  chomp will only remove the newline rather than any last character in the line.  This fixes a potential bug for the last message in the file if the file doesn't end in a newline.

### DIFF
--- a/scripts/quitrand.pl
+++ b/scripts/quitrand.pl
@@ -20,21 +20,17 @@ my $quitfile = glob "~/.irssi/irssi.quit";
 sub cmd_quit {
 	my ($data, $server, $channel) = @_;
 	
-	open(f,"<",$quitfile);
-	my @contenido = <f>;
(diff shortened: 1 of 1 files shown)


ailin-nemui:
it would seem that the array index can go out of the end of lines?

ailin-nemui:
@fpereda

fpereda:
It's been a while since I've used irssi (or IRC for that matter). I'm amazed someone is even caring about quitrand.pl at all!

vague666:
No idea if anyone uses it but it came up when someone wanted to script separate quit messages for each network :)
The perl code was quite outdated - the same with quitmsg.pl - so both scripts got a new shiny coat in case someone needs to use them

PR #720, opened 2020-08-08
In channels with a discord bridge, turns "<bridge> <Sender> Message" …

snan:
…into "&lt;Sender> Message", and hides spoilers.

### DIFF
--- a/scripts/discord_unbridge.pl
+++ b/scripts/discord_unbridge.pl
@@ -2,22 +2,19 @@ use strict;
 use warnings;
 use Irssi;
 
-our $VERSION = '1.2';
+our $VERSION = '1.3';
 our %IRSSI = (
     authors     => 'Idiomdrottning',
     contact     => 'sandra.snan@idiomdrottning.org',
     name        => 'discord_unbridge.pl',
-    description => 'In channels with a discord birdge, turns "<bridge> <Sender> Message" into "<Sender> Message", and hides spoilers.',
+    description => 'In channels with a discord bridge, turns "<bridge> <Sender> Message" into "<Sender> Message", and hides spoilers.',
     license     => 'Public Domain',
     url         => 'https://idiomdrottning.org/discord_unbridge.pl',
 );
 
 # HOWTO:
 #
-# optionally, but for better security, change
-#    if ($text) {
-# to
-#    if ($text && "Your Bridge's name goes here" eq $nick) {
+# set $bridgename to your bot's name, default is Yoda50.
 #
 # Regardless, to use the script just
 #   /load discord_unbridge.pl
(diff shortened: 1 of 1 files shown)


vague666:
Irssi can get confused by you changing the nick like that
A better way would be to rewrite the raw message. The following has been written for trigger.pl
`Soliton ┃ xacktm: this is exactly what i'm using, perhaps easier to adapt from that: 
/trigger add -rawin -regexp ':discordbot[^!]*!~discordbo@baldras.wesnoth.org PRIVMSG #(wesnoth.*) :<([^>]+)> (.*)' -replace ':d|$2!~Soliton@wesnoth/developer/soliton PRIVMSG #$1 :$3'`

snan:
Great idea!♥  
Pushed.

snan:
Hold on, with that implementation it breaks when people on Discord have spaces in their display names.

soliton-:
You can tweak the regex to allow spaces in nicks but I've never tested that. I've read what vague said as well that irssi can get confused if you force feed it invalid nicks. Might be worthwhile to just try and see what really happens though.

scripts/discord_unbridge.pl
```diff
@@ -0,0 +1,39 @@
...
+#
+# Based on discord_unhilight by Christoffer Holmberg, in turn
+# based on slack_strip_auto_cc.pl by Ævar Arnfjörð Bjarmason.
+
+my $bridgename = "Yoda50";
+
+sub bot_nick_change {
+    my($server, $message) = @_;
+    $message =~ s/:$bridgename!~[^ ]* PRIVMSG ([&#][^ ]+) :<([^>]+)> (.*)/:$2!~Yoda50\@ToscheStation PRIVMSG $1 :$3/;
+    $message =~ s/\|\|([^|][^|]*)\|\|/1,1$1/;
```
soliton-, line 35: I'm guessing the `1,1` changes the color. I think you may want to reset the color at the end and use the g flag to replace multiple spoilers in one message.

You can also simplify `([^|][^|]*)` to `([^|]+)`.

snan:
The regex in that `'server incoming'` version I pushed a few versions ago already allows the space in the nick in signal it continues. The problem is further down the stream from that.

scripts/discord_unbridge.pl
snan: I already do reset the color. The g and + are good ideas, thank you for those.

snan:
`<([^>]+)>` This is what allows spaces in Discord display names, both of our versions have that. One of my users find messages from these "space-heads" in their server log.

soliton-:
To clarify what I'm personally using now with trigger.pl to match the nick is ``<([][[:alpha:]_{}^`|][][[:alnum:]_{}^`|\-]*)>``. But as mentioned I have no idea what bad things actually do happen with invalid IRC nicks.

vague666:
spaces in a nick could segfault irssi but it's not something I've seen, mainly because I don't use discord nor discord with irssi

snan:
I hate Discord.

ailin-nemui:
maybe replace spaces, @ sign and exclamation mark with underscores if you want to cram them in the server incoming?

soliton-:
The issue with replacing characters is that you may get false duplicates then which could be confusing. The chance of that might be so low that it can be considered acceptable though. Also I'm not sure if you can get confusing duplicates anyway when people are allowed to change their nick on the discord side...

Another issue is not knowing how to highlight (`@nick`) discord users then but I believe highlighting discord users with spaces in their nick at least does not work anyway.

Given that there probably is no perfect solution for all these issues a solution that works well enough most of the time is what you should aim for and document the shortcomings.

snan:
I'm not sure what the shortcomings are, with the 'event privmsg'
approach that I original took, and then later reverted to when the
'server incoming' version broke for my users. Not saying there aren't
any, just that I haven't found them.

Ideally, I would want it to work on the display level, enough so you
could highlight, ignore etc the Discord users separately (instead of the
bridge as a whole), but not get confused into trying to /query them.

vague666:
Maybe @ailin-nemui can correct me if I'm wrong when I said irssi can get confused when you changed the nick the way you did in the 'event privmsg' handler

ailin-nemui:
@snan are you happy with the script for now?

snan:
Yes, I haven't experienced any problems so far, though I haven't
deliberately tried to stress test it with weird situations.

PR #722, opened 2020-08-21
bansearch: various fixes and improvements

dwfreed:
* Adjust set-by-server check to not be freenode-specific
* Make $~a check work
* Add $a check
* Make $x check work for all cases
* Shift mode redirect to block where mode request is sent
* Add debug output and setting to control it

### DIFF
--- a/scripts/bansearch.pl
+++ b/scripts/bansearch.pl
@@ -14,7 +14,7 @@ $VERSION = "1.3";
     license     => 'GPLv3+',
 );
 
-my($channel,$person,$nick,$user,$host,$real,$account,$string,$issues,$running,@jchannels,@jchannelstocheck);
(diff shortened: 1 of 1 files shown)


scripts/bansearch.pl
```diff
@@ -138,6 +138,17 @@ sub RPL_BANLIST {
 	}
 	# any logged-in user
 	if($maskreg=~m/^\$a$/i) {
+	    if($account!~m/^0$/) {
+		Irssi::active_win()->print(
+		    "$type against \x02$mask\x02 in $banchannel matches identified user." . $setby);
+		$issues++;
+	    }
+	    else {
```
vague666, line 146: Is this else needed since it does nothing?

vague666 (commented):
Maybe all the commented out else blocks can be removed or add a verbose setting so the prints in the else blocks can be used?

vague666 (approved):

PR #724, opened 2020-08-30
twitch_notify: init at 0.0.1

leocp1:
New script to print to the status window when a twitch channel has come online/offline.


ailin-nemui:
please explicityly `use POSIX`

PR #749, opened 2021-01-13
[CI] populate 'selfcheckcmd'

bw1:
cycle

### DIFF
--- a/_testing/config.yml
+++ b/_testing/config.yml
@@ -17,6 +17,7 @@ scripts_yaml_keys:
   - modified
   # special
   - HIDDEN
+  - selfcheckcmd
 cpan:
   broken_tests:
     - Geo::Weather


ailin-nemui:
why do you need this in the scripts.yml?

bw1:
for find scripts with a selfcheck enviornment

ailin-nemui:
@bw1 if you run this on a git clone of the scripts, you could just run it on every script. Originally at least, scripts.yml was meant for scriptassist

@irssi/developers anyone has an idea about this?

I don't like the design very much. Also the scripts are calling `Irssi::command("selfcheckhelperscript $result")`. I'm not sure if I like that either.

But I do like the idea of testing scripts!

vague666:
I'd prefer making sure the script is loaded and then calling the symbol directly, ie 
```
my $f = Irssi::Script::selfcheckscript->can('selfcheckhelperscript');
$f->($args) if $f;
```

bw1:
*  run it on every script 

that's the hard way and we hide the info

ailin-nemui:
Or maybe, just `Irssi::signal_add("print text"` in the selfcheckhelperscript

then `print "scriptname: self check result: $some_result"`;

that way you don't have to call a function.

ailin-nemui:
to find the scripts without putting too much info into the public scripts.yml, you can also try this: I think it might work (untested)

```
git clone -b ci-artefacts git@github.com:irssi/scripts.irssi.org scripts-ci
cd scripts-ci
perl \
    -MFile::Basename \
    -MYAML=LoadFile \
    -E'
        @docs = map {
            my ($doc) = @{ LoadFile($_) };
            $doc->{filename} = basename(dirname($_));
            $doc
        } @ARGV;
        for (@docs) {
            say $_->{filename} 
                if exists $_->{selfcheckcmd};
        }
    ' \
    -- \
    Test/*/info.yml
```

bw1:
or
```
	my $ft= read_text("$p/$fn",":latin1");
	if ( $ft =~ m/selfcheckcmd/ ) {
```
[:link:->:scroll:](https://github.com/bw1/check-irssi-scripts-selftest/blob/master/ModScanner.pl#L48)

bw1:
##  self check end :football:
```
 my $schs_version = $Irssi::Script::selfcheckhelperscript::VERSION;
 Irssi::command("selfcheckhelperscript $s") if ( defined $schs_version );
```
or

```
my $schs_version = $Irssi::Script::selfcheckhelperscript::VERSION;
Irssi::signal_emit("selfcheckend","ok") if ( defined $schs_version );
```
or

...

ailin-nemui:
* you cannot use signal_emit without signal_register in all scripts, so I would not use your idea (2)

* what do you think about vague's variant?

* If you use the "print 'special text'" idea, you do not need "IF / VERSION"

bw1:
### self check end
#### old
```    
Irssi::command("selfcheckhelperscript $s");
```
```
22:22 Users min:129 max:1879 check
22:22 -!- Irssi: Unknown command: selfcheckhelperscript
```
#### in the moment
```
my $schs_version = $Irssi::Script::selfcheckhelperscript::VERSION;
Irssi::command("selfcheckhelperscript $s") if ( defined $schs_version );
```
hint by @vague666 

#### version vague
```
my $f = Irssi::Script::selfcheckscript->can('selfcheckhelperscript');
$f->($args) if $f;
```

also two lines and a not marked  jump in

PR #759, opened 2021-01-31
[lastfm] v5.9

bw1:
@simmel


simmel:
Thanks for keeping it updated @bw1! LGTM :shipit:

PR #763, opened 2021-02-21
[urlgrab] improve url log file format

tscherf:
Added timestamp to urllog file.

### DIFF
--- a/scripts/urlgrab.pl
+++ b/scripts/urlgrab.pl
@@ -7,12 +7,13 @@
 #
 
 use strict;
+use Time::Piece;
 use Irssi 20010120.0250 ();
 use vars qw($VERSION %IRSSI);
-$VERSION = "0.4";
+$VERSION = "0.5";
 %IRSSI = (
-    authors     => 'David Leadbeater',
-    contact     => 'dgl@dgl.cx',
+    authors     => 'David Leadbeater, Thorsten Scherf',
+    contact     => 'dgl@dgl.cx, tscherf@redhat.com',
     name        => 'urlgrab',
     description => 'Captures urls said in channel and private messages and saves them to a file, also adds a /url command which loads the last said url into a browser.',
     license     => 'GNU GPLv2 or later',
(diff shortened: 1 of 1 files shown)


scripts/urlgrab.pl
```diff
@@ -7,12 +7,15 @@
...
-$VERSION = "0.4";
+$VERSION = "0.5";
 %IRSSI = (
-    authors     => 'David Leadbeater',
-    contact     => 'dgl@dgl.cx',
+    original_authors     => 'David Leadbeater',
+    original_authorscontact     => 'dgl@dgl.cx',
+    original    =>  'Thorsten Scherf',
+    original    =>  'tscherf@redhat.com',
     name        => 'urlgrab',
```
ailin-nemui, line 19: just add yourself to the authors list with a `,`, don't add any weirdness here

scripts/urlgrab.pl
```diff
@@ -55,11 +58,12 @@ sub find_url {
 }
 
 sub url_log{
+   my $dt3 = DateTime->now(time_zone => 'local');
```
ailin-nemui, line 61: DateTime is a heavy dependency, maybe you could port it to Time::Piece instead?

ailin-nemui:
@dgl

scripts/urlgrab.pl
```diff
@@ -55,11 +56,12 @@ sub find_url {
 }
 
 sub url_log{
+   my $t = localtime;
    my($where,$channel,$url) = @_;
    return if lc $url eq lc $lasturl; # a tiny bit of protection from spam/flood
    $lasturl = $url;
    open(URLLOG, ">>", $file) or return;
-   print URLLOG time." $where $channel $lasturl\n";
+   print URLLOG $t->datetime . " $channel $where $lasturl\n";
```
dgl, line 64: I guess I'm probably the only person who has anything to read this format, easy enough to deal with datetime changing and probably better.

Could we not swap `$channel` and `$where` randomly though, please.

tscherf:
@dgl

PR #767, opened 2021-03-08
[securemsg] add a new /sm rejidx and /sm showall command

ailin-nemui:


ailin-nemui:
@vague666

vague666:
Looks good to me :) I really should rewrite the script, it's a hodgepodge of horridness

PR #778, opened 2021-09-03
[trigger.pl] Sort trigger filters prior to save for idempotency

madduck:
For those of us keeping config in Git, it's a bit annoying when trigger changes the order based on the time of day and moon phase (i.e. hash order is undefined) on each config save, so let's sort the lines to make it a bit more idempotent.


ailin-nemui:
@coekie

ailin-nemui:
please increase the script version

madduck:
> please increase the script version

Done.

PR #781, opened 2021-09-04
[adv_windowlist.pl] Enable matching the F-keys as window changing shortcuts

madduck:
This enhances the AWL keymap parser to also match the use of F-keys
(i.e. F1 through F12), which will have to be mapped from their escape
sequences first (e.g. /bind ^[[11~ key F1).

The convention is followed that "meta-F1" gets displayed as "F1", and
"^F1" is shown as "^F1". Since it also makes sense to use F-keys without
modifiers, the syntax "`F1" is used to identify a "pure" F-key.

### DIFF
--- a/scripts/adv_windowlist.pl
+++ b/scripts/adv_windowlist.pl
@@ -503,15 +503,26 @@ sub _add_map {
 sub get_keymap {
     my ($textDest, undef, $cont_stripped) = @_;
     if ($textDest->{level} == 524288 and $textDest->{target} eq '' and !defined $textDest->{server}) {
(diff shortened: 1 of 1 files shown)


ailin-nemui:
It's not possible to use ^F1 in /bind to designate Ctrl+F-keys as far as I know

madduck:
> It's not possible to use ^F1 in /bind to designate Ctrl+F-keys as far as I know

That is true! But isn't it common to map the control sequences to e.g. `key F1` to keep the bindings clearer?

Would it harm to accept this?

PR #782, opened 2021-09-04
[ctrlact.pl] Show the window refnum in debug output if there's not window name

madduck:


ailin-nemui:
please update your version number...

PR #783, opened 2021-09-04
[adv_windowlist.pl] Match against /go argument when determining $Q

madduck:
I've recently started using `go2.pl` (a drop-in replacement to `go.pl`),
which provides the handy `/go` command to switch windows. Consequently,
I can now do:

```
/bind meta-s-meta-i command go #irssi
```

to switch to the window containing the best match for item `#irssi`.

This patch teaches AWL to read the `/bind` output for such keybindings
as well, so that `$Q` works accordingly.

### DIFF
--- a/scripts/adv_windowlist.pl
+++ b/scripts/adv_windowlist.pl
@@ -520,7 +520,7 @@ sub get_keymap {
 		if (/^change_window (\d+)/i) {
 		    _add_map(\%nummap, $1, $map);
 		}
-		elsif (/^(?:command window goto|change_window) (\S+)/i) {
+		elsif (/^(?:command (?:window goto|go)|change_window) (\S+)/i) {
 		    my $window = $1;
 		    if ($window !~ /\D/) {
 			_add_map(\%nummap, $window, $map);


ailin-nemui:
ok but why not just bind window goto, which is already supported?

madduck:
@ailin-nemui honestly, I don't know. When I added `/go` to my configuration in oh-my-goodness-i-do-no-remember-how-long-ago, I don't think `/window goto` existed. Now, the only thing `/go` seems to have over `/window goto` is that it takes the `chatnet` parameter too, which helps in cases where you've joined the same channel on multiple networks. I only have 2 of those instances left, so maybe it's time to move on? On the other hand, does it hurt to add support, given how simple the patch is?

ailin-nemui:
my concern is that /go may be using partial matching or another algorithm than Irssi, thus the displayed key binding could be incorrect

madduck:
Since I've found out that `window goto` will simply cycle through window items that share the same name across different chatnets. It still does not allow me to deterministically go to one specific window item.

Instead of hard-coding `go.pl`, I suggest that I submit a patch that makes this configurable, default off (and leave `window goto` always enabled).

PR #799, opened 2022-01-04
[leodict] Add proxy support for requests made to dict.leo.org

morgenm:
Minor addition to allow users to set and use proxies for leodict.pl's requests to the leo website. Helpful for certain setups and networks where proxies are required, but not all irssi traffic may be easily tunneled.

### DIFF
--- a/scripts/leodict.pl
+++ b/scripts/leodict.pl
@@ -4,7 +4,7 @@
 use strict;
 
 use vars qw($VERSION %IRSSI);
-$VERSION = '20210126';
+$VERSION = '20220104';
 %IRSSI = (
     authors     => 'Stefan \'tommie\' Tomanek, bw1',
     contact     => 'bw1@aol.at',
@@ -347,7 +347,6 @@ my %options = (
 sub cmd_leodict ($$$) {
     my ($args, $server, $witem) = @_;
     utf8::decode($args);
-    #my $burl= "https://dict.leo.org/";
     my $burl = "https://dict.leo.org/";
     my $url;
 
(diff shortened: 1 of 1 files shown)


scripts/leodict.pl
```diff
@@ -175,6 +200,22 @@ ($$$)
...
+	if ($proxy_type ne 'socks' && $proxy_type ne 'https' && $proxy_type ne 'http') {
+	    print CLIENTCRAP "%R>>%n Invalid proxy type: $proxy_type.";
+	    return;
+	}
+	if ($proxy_port eq 0) {
+	    print CLIENTCRAP "%R>>%n Please specify a proxy port.";
+	    return;
+	}
+    }
+
```
bw1, line 218: there is a better place in `sub sig_setup_changed`

bw1 (commented):
nice :)

scripts/leodict.pl
```diff
@@ -380,6 +424,9 @@ sub sig_setup_changed {
 Irssi::settings_add_str($IRSSI{'name'}, 'leodict_default_options', '-en -both');
 Irssi::settings_add_int($IRSSI{'name'}, 'leodict_paste_max_translations', 2);
 Irssi::settings_add_bool($IRSSI{'name'}, 'leodict_paste_beautify', 1);
+Irssi::settings_add_str($IRSSI{'name'}, 'leodict_http_proxy_address', 'none');
```
vague666, line 427: Instead of using `'none'` and forcing the user to possibly remember to use it to unset the proxy type, use `''` or `undef` and check for corresponding value in the `if statements` above

PR #801, opened 2022-01-19
[tab_stop.pl] Implement tab stops with configurable length

kjetilho:
This changes the default behaviour to make TAB replacements line up. To get the original behaviour, the user will need to do `/set
tabstop_interval 0` explicitly.

### DIFF
--- a/scripts/tab_stop.pl
+++ b/scripts/tab_stop.pl
@@ -9,28 +9,38 @@
 #
 # 24.05.2011
 # * Buggered about with by shabble.
+#
+# 19.01.2022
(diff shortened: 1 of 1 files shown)


ailin-nemui:
@wertarbyte @shabble

kjetilho:
It is possible the default should be 0 for backwards compatibility, and also use tabstop_replacement rather than hardcoding SPC when making the tabstop.  e.g., if someone wants to distinguish TABs from SPC using a sequence of '…' or some other character.  That would not be so useful for cut-n-paste, of course.

PR #804, opened 2022-02-28
[elist.pl] added option to remove colors from topics

aenigma-es:
Added an option to have the channels list printed with colors stripped from topic text.
use: "/set elist_colorized ON/OFF" to change behavour (default is ON)


ailin-nemui:
@icassina

PR #805, opened 2022-03-04
[intercept.pl] detect Irssi commands as first word

vague666:


bw1 (changes requested):
merge

vague666:
> mismatch is a `merge`
not sure I understand, or rather, I am sure I don't understand :) What should change?

PR #808, opened 2022-04-27
[colorize_nicks] Case sensitive selection

capitaineplanete:
make colorise_nicks case sensitive

### DIFF
--- a/scripts/colorize_nicks.pl
+++ b/scripts/colorize_nicks.pl
@@ -1,7 +1,7 @@
 use strict;
 use warnings;
 
-our $VERSION = '0.4.1'; # ed9cb119fc4b3d1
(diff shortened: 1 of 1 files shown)


ailin-nemui:
why?

capitaineplanete:
so may false positives, at least in my experience. Besides most if not all clients match nick case in tab completion.

I mean look at this... :P most of which are the same word repeated.

Screenshot 2022-04-27 at 10 35 52

ailin-nemui:
from my understanding your patch will not change anything. but maybe I'm wrong

capitaineplanete:
you're right my tests weren't as thorough. I'm making a few changes and will add an option to opt in. I don't really perl I'm just following the control flow of the code

PR #810, opened 2022-05-24
array returns from glob and wrong filehandle

aenigma-es:
Patched the code to solve two little bugs:
- corrected glob as it returns an array and not a single value
- corrected wrong filehandle name when closing file


alepore (approved):
LOL! This is ancient

PR #818, opened 2022-07-29
Add yubikey.pl script

ppaeps:
This script helps prevent accidentally sending YubiKey OTPs to IRC.

I wrote this many years ago.  Sharing it with the wider community after sharing it with another person whose cat likes to play with their YubiKey. :)

### DIFF
--- a/scripts/yubikey.pl
+++ b/scripts/yubikey.pl
@@ -1,0 +1,21 @@
+#!/usr/bin/env perl
+
+use strict;
+use vars qw($VERSION %IRSSI);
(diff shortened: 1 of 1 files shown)


vague666:
Maybe rename the script? Now it sounds like it adds yubikey support, so stop_yubikey.pl or similar might be more descriptive
My 2c :)

ppaeps:
Thanks @vague666.  Good idea.  I went with an even clearer filename.  
The script doesn't try to stop anyone from using a YubiKey either. :-)

PR #821, opened 2022-09-04
add topic-diff.pl

bw1:


JustAnotherArchivist:
Just for the record because I was searching for the origins of the script earlier: it's not entirely clear whether Pascal Hakim (@pasc) authored the entire file. This is coming from the Debian package irssi-scripts, which [imported it](https://salsa.debian.org/debian/irssi-scripts/-/commit/00d11ffa30df5f0421bfd556436d60c22ef4c1da) from [an SVN repository](https://sources.debian.org/src/irssi-scripts/20220704/INDEX/#L1465) of Christoph Berg (@df7cb) in 2004 (note: not the date of that svn-inject commit, cf. `debian/changelog` file there). In 2013, [a small change](https://salsa.debian.org/debian/irssi-scripts/-/commit/715162a92cbffaadd0f917c3a259edb1441ccc5a) to the script was committed with the message 'New upstream release', suggesting a commit on said SVN repository, possibly by Christoph. (Whether the original import was as written by Pascal is of course also not confirmed.) The SVN repo is a 404 these days, so easy verification isn't possible.

df7cb:
It's been a decade or two since I stopped maintaining the irssi-scripts package in Debian, so I forgot most of the details already. But I can probably say pretty confidently that I was never upstream of topic-diff.pl, if that was the question.

pasc:
I've got no recollection of whether anyone else contributed to this file. I definitely wrote the original version as there was nothing else like that at the time that I could find. Hope that helps!

PR #822, opened 2022-09-10
add url.pl

bw1:


bw1:
debian 
https://sources.debian.org/data/main/i/irssi-scripts/20220704/INDEX

PR #827, opened 2022-11-21
apply consistent tabs as spaces

zdleaf:


scripts/urlwindow.pl
```diff
@@ -7,14 +7,13 @@
...
 
-$VERSION = "1.3";
+$VERSION = "1.4";
 %IRSSI = (
     authors     => "zdleaf",
     contact     => 'zdleaf@zinc.london',
     name        => "urlwindow",
     description => "Log all urls from #channels and /msgs in a separate window",
     license     => "Public Domain",
-    url         => "http://irssi.org/",
```
bw1, line 17: tab?

zdleaf: Not a tab, just didn't think removing this warranted a separate commit. Could have mentioned it in the commit message perhaps.

zdleaf: Can make a separate commit or add something to commit message about removing redundant/irrelevant irssi.org URL?

PR #829, opened 2022-12-01
Add /cubes_text to show colors applied to the text, not the background

kot0dama:
Helpful for people who want to select colors for texts, based on their current background.


hloeung:
Very handy!

PR #834, opened 2022-12-28
trigger: match irc op_public as publics

dwfreed:
These are often called statusmsg or wallchops, and are often also used
by channel modes that cause messages that would otherwise be blocked to
be sent to chanops.

Adds a prefix filter to allow differentiating these from regular
publics.

Also extends the debug output to include the trigger that tripped the
debug


dwfreed:
Last commit fixes an issue where the expands make it into the debug output unnecessarily

PR #838, opened 2023-03-03
yubikey: update regex

Kakama:
Old script behaviour: match on 6 c characters
Yubikey OTP format: 4 c followed by 40 other modhex characters
New script regex matches better for less public shame with yubikey nano-c

### DIFF
--- a/scripts/suppress_yubikey_otp.pl
+++ b/scripts/suppress_yubikey_otp.pl
@@ -4,7 +4,7 @@ use strict;
 use vars qw($VERSION %IRSSI);
 use Irssi;
 
-$VERSION = "20151026";
+$VERSION = "20230303";
 %IRSSI = (
     authors     => "Philip Paeps",
     contact     => "philip\@trouble.is",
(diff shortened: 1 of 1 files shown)


scripts/suppress_yubikey_otp.pl
```diff
@@ -16,6 +16,7 @@
 
 sub event_send_text {
     my ($line, $server_rec, $wi_item_rec) = @_;
-    Irssi::signal_stop() if $line =~ /^cccccc/;
+    Irssi::signal_stop() if $line =~ /^cccc[bcdefghijklnrtuv]{40}/;
```
dwfreed, line 19: Would be nice if this also matched the custom OTPs, which always start with 'vv'

dwfreed (changes requested):
I can't comment on it inline, but `VERSION` always needs to be increased when submitting a PR.

scripts/suppress_yubikey_otp.pl
dwfreed:
```suggestion
    Irssi::signal_stop() if $line =~ /^(cc|vv)[bcdefghijklnrtuv]{42}/;
```

Kakama: Thank you, I used today's date for the VERSION now and added your changes for the custom OTPs.

dwfreed (approved):

PR #841, opened 2023-04-16
mail.pl: show correct option in manual

abbbi:
hi,

mail.pl refers to /SET MAILDIRMODE which is wrong, the script registers option maildir_mode
commit fixes this.

### DIFF
--- a/scripts/mail.pl
+++ b/scripts/mail.pl
@@ -1,6 +1,6 @@
 use strict;
 use vars qw($VERSION %IRSSI);
-$VERSION = "2.92";
+$VERSION = "2.93";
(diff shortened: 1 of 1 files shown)


scripts/mail.pl
bw1, line 1: a new version per pull request please [line 3](https://github.com/irssi/scripts.irssi.org/pull/841/files#diff-695197ba3076543f1f381fa08613a7898b9dce390e354b445f7084039baba0b1R3)

bw1 (changes requested):

bw1 (approved):

PR #849, opened 2023-07-02
added the new script, whereami, queries for the IP address and displays that in a statusbar item

terminaldweller:
whereami queries an IP service which returns the current IP and adds that as a statusbar item. the default URL uses icanhazip.com.
I have added a note regarding that service being operated by cloudflare for anyone that cares. the url can be replaced through settings.

### DIFF
--- a/scripts/whereami.pl
+++ b/scripts/whereami.pl
@@ -1,0 +1,50 @@
+use Irssi;
+use LWP::UserAgent;
+use strict;
+
+our $VERSION = "0.1";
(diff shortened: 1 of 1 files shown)


terminaldweller:
thanks Nei.

vague666:
Maybe add a setting(so user can decide if they want to do it) to update dcc_own_ip if the returned ip doesn't match the setting?

terminaldweller:
sure. i could add that in and set it to off by default.

PR #853, opened 2023-08-21
added a blacklist option, added host and port as new options

terminaldweller:
Like the title says, added a blacklist, host and port option.
tested with postgres 15.4 and 14.2 .

### DIFF
--- a/scripts/irssi_logger.pl
+++ b/scripts/irssi_logger.pl
@@ -23,6 +23,7 @@ $VERSION = "1.0";
 
 my $user = $ENV{LOGNAME} || $ENV{USER} || getpwuid($<);
 my $dbh;
+my %blacklist = ();
 
 
 my $sql = qq~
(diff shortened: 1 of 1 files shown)


terminaldweller:
i dunno how. lemme go read another script and fix that.
and yes. i should increase the version too.

scripts/irssi_logger.pl
```diff
@@ -90,29 +93,35 @@ sub connect_db {
...
 
-    return DBI->connect("dbi:Pg:dbname=$dbname", $dbuser, $dbpass) || Irssi::print("Can't connect to db!" . DBI::errstr);
+    return DBI->connect("DBI:Pg:database=$dbname;host=$dbhost;port=$dbport", $dbuser, $dbpass) || Irssi::print("Can't connect to db!" . DBI::errstr);
 }
 
 sub write_db {
     my ($nick, $message, $target) = @_;
-    my @vals;
-    my $date = strftime("%Y-%m-%d %H:%M:%S", localtime);
+    if (exists $blacklist{$target}) {
```
qbit, line 108: Personally, i'd just do an `if (!exists` here and exclude the `} else {`.

qbit (approved):
LGTM

scripts/irssi_logger.pl
terminaldweller: sure

PR #854, opened 2023-08-27
Add zmartfilter.pl script

zelgomer:
smartfilter.pl reimagined, optimized for unusually flakey networks such as IRC over I2P


zelgomer:
Yes, I wrote it.  However, this is an anonymous identity that I use on the I2P network, so assigning credit to this name does not have much meaning.  For this reason, I opted to release it to the public domain with no credit.

PR #856, opened 2023-09-10
experimental recognition of python scripts

ailin-nemui:


dwfreed:
How does the website handle python scripts? I would not want them mingled with perl scripts; probably a mode switcher at the top to select between perl and python would work best, along with a warning that python is not going to work for most people at the moment.

dwfreed:
I absolutely expect users to not read the column or the warning message, and then complain when they can't load python scripts in their regular (perl) irssi.  Having a mode switcher that hides the other language makes things easier in general for both language users. Eg, once this is merged and we start putting python scripts in the repo, the python users won't have to search through all the perl scripts to find the python ones.

dwfreed:
Then perhaps a toggle that controls whether python scripts are displayed?

(For some reason I was thinking irssi-python was a full patched irssi that *replaced* perl with python, not *just* a module for an existing irssi also capable of perl)

Until the python module is widely available to distro users, and perhaps there's an implementation in core to detect mismatched command vs script language, I still foresee that mingling the python scripts in with the perl scripts is going to cause lots of headache for users and support.

PR #865, opened 2023-11-28
add trivia.pl script

terminaldweller:
it accepts a list in the form of "server1/#channel1,Tag1 regex2,Tag2" and regex-substring-matches againt the server and window name. it it matches, returns the tag for the winsow as an expando.
I use it to orgaize different protocols. it helps to see a matrix or xmpp tag somewhere if you have almost 200 windows.

### DIFF
--- a/scripts/trivia.pl
+++ b/scripts/trivia.pl
@@ -1,7 +1,6 @@
 use Irssi;
 use strict;
 use warnings;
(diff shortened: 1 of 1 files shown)


terminaldweller:
The test is complaining about the `use v5.36`. should I remove it?

PR #867, opened 2023-12-29
added speedread.pl

nauski:
Added speedread.pl. The script bolds the first (1-3) characters of each incoming or outgoing word to make reading easier for people with dyslexia or other reading impairments.


nauski:
Ouch good catch, should be visible locally only now.

PR #868, opened 2023-12-30
added mqtt.py, sends your private messages over mqtt

terminaldweller:
sends your private messages over mqtt.
requires the paho python client.


terminaldweller:
This one fails the test because it needs a dependency from pip iirc.

PR #875, opened 2024-05-07
added a new script dont_reconnect.pl

terminaldweller:
A simple script. takes in a space-separated list of server names. runs rmreconn after each server in the list reconnects.
If more than one server reconnects and one of them is not on the list, well too bad. The script will disable reconnection for that as well.


terminaldweller:
I'm closing this due to a lack of interest from the one person this was supposed to help.

PR #879, opened 2024-06-21
Irssi logger - reconnect to db and handle blacklist param change

terminaldweller:
this includes two changes:
* now on checking for error while trying to commit a row, if we cant ping the db, we reconnect
* added a function to handle the change for the blacklist parameter


terminaldweller:
closing for obvious reason

PR #885, opened 2025-01-04
spelling error

jwlynch:
on line 150, include is misspelled as inlcude


jwlynch:
the.... script? is it that the line I replaced the word on, is the script
yyou're referring to? I'm not (yet) an irssi user, and I don't know
how your version numbers work. First, where is the version number I would
replace? Second, how exactly do I bump it?

=Jim

PR #890, opened 2025-08-16
cron: Fix time drifting bug.

RealKindOne:
Currently the code does this:
1. Get time
2. Start `sig_timeout` at next `hh:mm:00`
3. Re-run `sig_timeout` every 60 seconds. // This is the bug since it takes extra time (milliseconds) to run the command. That eventually adds up to seconds, then minutes...

My changes:

Re-run `sig_timeout` at every `hh:mm:00`

### DIFF
--- a/scripts/cron.pl
+++ b/scripts/cron.pl
@@ -75,7 +75,7 @@ use Irssi;
 use strict;
 use vars qw($VERSION %IRSSI);
 
-$VERSION = "0.12";
+$VERSION = "0.13";
 %IRSSI = (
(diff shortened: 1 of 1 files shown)


cvbge:
Hi :)

Now this is a surprise :)

I haven't used this script (or IRC/Irssi) for the last 20 years. Nice to see someone still uses the script, I forgot about it totally. 

I'm not sure I can help with this pull request, I don't remember the code, Irssi API, or even perl that much...

RealKindOne:
@ailin-nemui fixed

PR #905, opened 2026-03-30
Add mention_report.pl to cache and report mentions

Exaga:
The script mention_report.pl acts as a silent buffer for your irssi terminal. It eliminates the need to manually scroll and search through thousands of lines of channel noise to find out if someone actually wanted you. 

It's specifically designed to work alongside autoaway.pl to solve the "missed mentions" problem.

A functional breakdown of what it does while you are away:

1. The Trap (while you're away)

When you type /AWAY [reason], the script initialises a temporary memory cache (an array).

    It ignores regular chatter.

    It intercepts any line where your nick is mentioned or any Private Message (Query).

    It strips the formatting and timestamps each hit, storing it in RAM.

2. The Dump (when you return)

When you type /away (to unset it) or /back, the script triggers and lists all messages containing your nick that were sent while you were away.

    Instead of you having to grep your ment.log or scroll up 500 pages, it replays those cached mentions directly into your active window.

    It presents them in a clean list: [Time] [Channel/Nick] <Message>.

3. The Flush

Once the report is printed to your screen, it clears the cache. This keeps your system memory usage near zero and ensures you don't see the same "old" mentions the next time you go away.

### DIFF
--- a/mention_report.pl
+++ b/mention_report.pl
@@ -1,0 +1,83 @@
+# mention_report.pl - Caches nick mentions while away and reports on return.
+#
+# Copyright (c) 2026 Exaga - SAIRPi Project : https://sairpi.penthux.net/
+# 
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


vague666:
All signs point to AI slop, won't merge

Exaga:
My most sincere apologies. Had I been aware that such work was not welcome I would not have wasted your time.

SAIRPi Project exists for research and development on AI Edge computing.

PR #907, opened 2026-06-16
twtopic: 2.00 update -- Add dynamic statusbar resizing to window width, foreground colouring, minor optimizations.

Erwin-Iosef:
This PR updates `twtopic.pl` to 2.00 with these features:

- The topic bar dynamically resizes with respect to the active window width, (thanks a lot to armin for the work!).
- New formatting option that lets users color the foreground.
- Some minor optimizations.
----
Since this script differs from the original a lot, I was unsure whether to release the modifications in a separate script instead of modifying this, if necessary I'm happy to submit a new script instead of modifying the old one.

### DIFF
--- a/scripts/twtopic.pl
+++ b/scripts/twtopic.pl
@@ -4,117 +4,134 @@ use Irssi;
 use Irssi::Irc;
 use Irssi::TextUI;
 
-$VERSION = '1.02';
+$VERSION = '2.00';
 %IRSSI = (
-   authors	=> 'John Engelbrecht',
-   contact	=> 'jengelbr@yahoo.com',
+   authors	=> 'John Engelbrecht - original script, dynamic resizing logic by Armin Jenewein, additional modifications by Erwin Iosef',
+   contact	=> 'a@m2m.pm, erwiniosef@gmail.com, jengelbr@yahoo.com',
    name	        => 'twtopic.pl',
-   description	=> 'Animated Topic bar.',
+   description	=> 'Animated Topic bar with dynamic resizing to window width.',
    sbitems      => 'twtopic',
    license	=> 'Public Domain',
-   changed	=> '2018-09-08',
+   changed	=> '2026-06-16',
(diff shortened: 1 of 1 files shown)


scripts/twtopic.pl
```diff
@@ -4,117 +4,135 @@
...
+my $start_pos;
+my $size;
+my $color;
+my $topic;
+my @mirc_color_arr = ("\0031","\0035","\0033","\0037","\0032","\0036","\00310","\0030","\00314","\0034","\0039","\0038","\00312","\00313","\00311","\00315","\017");
 
+# Setup timer
 sub setup {
    my $time = Irssi::settings_get_int('twtopic_refresh');
    Irssi::timeout_remove($timeout) if ($timeout != 0);
```
Erwin-Iosef, line 51: Not sure what to do here, adding line 57 before this breaks the script after one second, plus it's how it was originally.

scripts/twtopic.pl
```diff
@@ -4,117 +4,134 @@
...
-   $text="[".$text."]";
-   $item->default_handler($get_size_only,$text, undef, 1);
+
+# Status bar show handler
+sub show {
+   my ($item, $get_size_only) = @_;
+   $topic = get_topic($item);
+   $color = Irssi::current_theme->get_format(__PACKAGE__, 'twtopic_fgcolor');
+   #Add '[,]' to both ends of topic string if you want '[ topic ]', ie($color\[$topic\]%n)
+   $topic = " $color$topic%n";
```
Erwin-Iosef, line 67: The only missing piece is to let the user format the topic bar foreground how they want instead of just the color, like how I mentioned in L65 above. 

It's not a big deal since the user can do that anyway inside the file.

And the mechanism to add inside the file via a function seems too complicated/adding execution overhead for such a simple thing in the first place.

Erwin-Iosef:
It's ready for approval!

PR #910, opened 2026-08-05
[0x0st] remove

bw1:
server returns:
"uploads disabled because it's been almost nothing but AI botnet spam for the past few months. will be back with a few changes at some point. no ETA."


dgl:
Someone could use this with a different site, if it accepts post in the right way (and the 0x0st code is open, or it's trivial to make a clone). There are a few, although I won't name them here because they tend to get abused...

PR #912, opened 2026-08-14
[omdb] create omdb.pl

bw1:
new script for replace imdb.pl

### DIFF
--- a/scripts/omdb.pl
+++ b/scripts/omdb.pl
@@ -1,0 +1,333 @@
+use strict;
+use vars qw($VERSION %IRSSI);
+
+use Irssi;
+use File::Fetch;
+use File::Basename;
+use File::Path qw/remove_tree/;
+use JSON::PP;
+use Storable qw/store_fd fd_retrieve/;
+use Text::Wrap;
+
+$VERSION = '0.1';
+%IRSSI = (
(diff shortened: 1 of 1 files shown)


scripts/omdb.pl
```diff
@@ -0,0 +1,335 @@
+use strict;
```
vague666, line 1: Add use warnings;

scripts/omdb.pl
```diff
@@ -0,0 +1,335 @@
...
+		if ( $c < 10 ) {
+			$lastpage=0;
+		}
+		Irssi::print(draw_box($IRSSI{name}, $t,$c." of ". $r->{totalResults}, 1)
+				, MSGLEVEL_CLIENTCRAP);
+	}
+}
+
+sub cmd {
+    my ($args, $server, $witem)=@_;
```
vague666, line 222: This is my preference but I'd consider moving the 'my $url's below here and make it
`my $url="http://www.omdbapi.com/?apikey=".Irssi::settings_get_str('omdb_apikey');`
and then add the other path variables to $url
Add a `return unless length Irssi::settings_get_str('omdb_apikey') > 0;` also to safeguard against empty values

PR #914, opened 2026-08-15
[inputlength] Add feature to count backwards from the maximum possible length

kontaxis:
This change adds boolean setting `inputlength_countdown`. When `ON`, instead of counting the length of `inputline` up from zero, it counts down from the maximum length of a contiguous message per RFC 1459, i.e., 512 minus overhead.

The code dynamically determines the overhead given the current nickname, ~~userhost~~, and channel.

This is useful when the user wants avoid their message being split into two, e.g., because they are over the limit by a word.

Setting `inputlength_countdown` is `OFF` by default.

This change also adds an integer setting `inputlength_offset`. Its value is added to the actual length of `inputline`. This is useful when other settings affect the final length, e.g., `split_line_start` or `split_line_end`. In that case `inputlength_offset` should be set to account for those additional characters.

As an example, when `split_line_end = "> "` and `split_line_start = " >"` then `inputlength_offset = 4`.

Setting `inputlength_offset` is 0 by default.

### DIFF
--- a/scripts/il.pl
+++ b/scripts/il.pl
@@ -3,27 +3,27 @@
 #   inputlength = "{sb length: $@L}";
 #
 #  with leading spaces: (3 spaces in example)
-#    inputlength = "{sb $[-!3]@L}"; 
+#    inputlength = "{sb $[-!3]@L}";
 #
 #  with leading char "-"
 #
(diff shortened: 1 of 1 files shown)


bw1:
thx

